PR #12, opened 2016-07-26
Added easing transition on hovers

reenybeeny:
Added easing transition on hovers
Changed opacity on button hovers


gender404:
Gulp is automatically adding prefixing where needed so you don't need to enter them.

-webkit-transition: all 0.4s ease;
-moz-transition: all 0.4s ease;
-o-transition: all 0.4s ease;
transition: all 0.4s ease;

PR #27, opened 2016-08-09
Level0 - Page Title created

gender404:
Also merged in the MGRV-51 pull request currently open.


styleguide/source/_patterns/03-organisms/page-banner.twig
```diff
@@ -0,0 +1,21 @@
+<section class="page-banner">
+  <style>
+    .page-banner__background {
+      background-image: url('/assets/images/temp/1600x400.png');
```
jasonstanbery, line 4: @legostud

1) Is this better than just putting the style inline on the div?
2) Instead of adding the temp image, c(sh)ould we just use http://placehold.it/1600x400

Both of these are likely just personal preference, so if this is your preferred way I'm good with it too.

gender404:
1. the Style block makes it easier to switch backgrounds based on break points
2. I'm using temp images, because this is a living style guide.  As such, I'm not sure if placehold.it will continue to exist for the life of this style guide.

PR #31, opened 2016-08-11
Design QA updates

reenybeeny:
-padding on search button
-updated some styles in util nav
-updated an icon
-created a link transition mixing and applied to various places
-made some font-weight and link hover style changes to the utility nav


styleguide/source/assets/scss/02-molecules/_main-nav.scss
```diff
@@ -82,6 +82,7 @@
     @include ma-chevron;
     &:after {
       cursor: pointer;
+      margin-left:6px;
```
jasonstanbery: @reenybeeny I know, I know, I know - this is super duper nit picky and doesn't change the styles generated by the sass, but could you put a space between the colon and the value here, and the ones in settings_util.scss & utility-nav.scss

styleguide/source/assets/scss/02-molecules/_settings-util.scss
```diff
@@ -59,7 +59,7 @@
       }
 
       &__link {
-        opacity: .7;
+        font-size:16px;
```
jasonstanbery: I believe the base font-size is 16px, and we're using REMs as our unit of measurement for fonts. So this should be `font-size: 1rem;` instead of 16px.

styleguide/source/assets/scss/02-molecules/_dropdown.scss
```diff
@@ -30,11 +30,11 @@
   &__link {
     @include ma-chevron;
     display: inline-block;
-    letter-spacing: .075em;
     padding-right: .5em;
 
     &:after {
       margin-left: .5em;
+      opacity:.7;
```
jasonstanbery: @reenybeeny: uno mas - then it's golden.

PR #39, opened 2016-08-16
changed decorative border opacities and padding

reenybeeny:
changed decorative border opacities and padding
resized font on breadcrumbs, page banners, top tasks
updated font weights on items in the page banner

### DIFF
--- a/styleguide/source/assets/scss/00-base/mixins/_ma-border-decorative.scss
+++ b/styleguide/source/assets/scss/00-base/mixins/_ma-border-decorative.scss
@@ -16,7 +16,7 @@
     }
     @else {
       background-color: $fg;
-      background-color: rgba($fg,1);
+      background-color: rgba($fg,.5);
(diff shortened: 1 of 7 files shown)


styleguide/source/assets/scss/00-base/mixins/_ma-border-decorative.scss
```diff
@@ -16,10 +16,9 @@
     }
     @else {
       background-color: $fg;
-      background-color: rgba($fg,.5);
+      background-color: rgba($fg,1);
```
jasonstanbery: @reenybeeny 
We can actually just remove this line (#19) if we're setting it to 1 because it has the same result as line 18 (which is there to be a fallback for old IE).

gender404: I'm confused by this change.  line 18 is for IE8, line 19 is for IE9, and line 20 is for modern browsers.  So this change is only affecting IE9.

PR #55, opened 2016-08-30
Level2 - Phase 1

gender404:
1. Updated Callout Link molecule to use JSON and be more re-usable
2. Updated Page Header organism to use JSON data
3. Added a placeholder to the input atom
4. Created a new Keyword search molecule
5. Created a new Action Finder Organism
6. Created a new Level2 page in PatternLab


styleguide/source/_patterns/03-organisms/page-header.twig
```diff
@@ -1,4 +1,4 @@
 <section class="ma__page-header">
-  <h1>Get a State Park Pass</h1>
-  <div class="ma__page-header__intro"><p>Do you regularly visit Massachusetts parks or beaches? Our MassParks Pass offers you a significant savings over daily parking charges.</p></div>
+  <h1>{{pageHeader.title}}</h1>
+  <div class="ma__page-header__intro">{{pageHeader.intro | raw}}</div>
```
jasonstanbery, line 3: Can we try using the `escape` filter instead of raw? I assume when this goes into Palantir's build they'll update all of these with more than just the field name so it shouldn't be an issue, but just to be safe, if escape works we should use it.

http://twig.sensiolabs.org/doc/filters/escape.html

gender404: I just tried |escape, |e, |e('html') and none of those worked.

jasonstanbery: Hmm, okay.

PR #71, opened 2016-09-13
Fix up deployment errors

froboy:
This removes some extra spacing in circle.yml but primarily adds some lines which fix git errors related to automated deployments. @legostud please let me know if you'd like to review and merge. This should have no effect on your process, but I don't want to merge to your repo without review.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -13,8 +13,11 @@ test:
         - npm run gulp prod
 
 deployment:
+    pre:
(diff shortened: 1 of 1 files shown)


froboy:
Aha, circle still passes, but doesn't actually deploy (which is correct now). https://circleci.com/gh/massgov/mayflower/54 is a demonstration of a successful deploy.

PR #73, opened 2016-09-14
Home pilot

gender404:
1. New Chang Log organism
2. New Whats New organism
3. New 50/50 Split organism/template
4. Renamed underline-link mixin to link-underline
5. created a new link-decorative mixin
6. New Decorative Link atom (used in a lot of places)
7. Refactored the Related Actions organism to use the Decorative Link atom.


styleguide/source/_patterns/03-organisms/related-actions.json
```diff
@@ -0,0 +1,19 @@
+{
+  "relatedActions": {
+    "items": [{
+      "link" : {
+        "type": "",
+        "href": "#",
+        "text": "Give a Park Pass a as a Gift!"
```
jasonstanbery, line 7: I think you've got a typo here. I'm using it to test the "Code Review" thing, so deal with it.

jasonstanbery (approved):
Looks good, I'm going to merge this even with the typo because it's placeholder copy, but we might want to fix this since it reads like legit copy.

PR #101, opened 2016-09-26
Design QA updates

reenybeeny:
Updated some text 
Changed margin on smaller breakpoints for header on L.0 page
Updated font-size and margin on descriptions on L.0 page


styleguide/source/_patterns/03-organisms/by-author/top-actions.json
```diff
@@ -1,6 +1,6 @@
 {
   "topActions": {
-    "title": "Common Actions",
+    "title": "Common Actions &amp; Guides",
```
gender404: "title": "Common Actions &amp; Guides", = > different rules for html and json.  In json you can leave it as & instead of &amp;

PR #103, opened 2016-09-26
Design QA updates

reenybeeny:
-changed line height on H4's
-changed width of download icons on action page so there was less space between the icon and text
-updated padding on action page tags and added a hover state
-increased opacity on "GUIDE" text on L2 to meet accessibility requirements

### DIFF
--- a/styleguide/source/assets/scss/02-molecules/_header-tags.scss
+++ b/styleguide/source/assets/scss/02-molecules/_header-tags.scss
@@ -9,15 +9,15 @@
     a {
       border: 2px solid;
       display: block;
-      line-height: 2.29;
+      line-height: 2;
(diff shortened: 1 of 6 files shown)


styleguide/source/assets/scss/02-molecules/_header-tags.scss
```diff
@@ -9,19 +9,19 @@
     a {
       border: 2px solid;
       display: block;
-      line-height: 2.29;
+      line-height: 2;
       margin-bottom: .5em;
-      padding: 6px 8px 3px;
+      padding: 0 8px 0;
       text-transform: capitalize;
```
gender404, line 15: padding: 0 8px 0; => should be padding: 0 8px;

PR #104, opened 2016-09-27
Action page extras

gender404:
1. New Sequential List component for the Action page.
2. Modified the Action page Sections block to make the title and intro optional
3. Bug fixes
4. Removed the Piggy Bank icon
5. Fixed text wrapping issue with Link List component
6. Refactored the Pattern Lab flow for including Action Steps to remove an unneeded layer


styleguide/source/assets/scss/02-molecules/_action-seq-list.scss
```diff
@@ -0,0 +1,33 @@
+$action-seq-list-icon-border: 3px;
+.ma__action-seq-list {
+
+  &__items {
+    @include ma-reset-list;
+    counter-reset:li; // Initiate a counter
```
jasonstanbery: is there a reason why this rule, and the ones on line 17 & 18 don't have spaces before the values? Is this because of something with how the counter property behaves?

jasonstanbery (commented):
Looks good, just a small question - totally merge-able as is, call it a curiosity.

styleguide/source/assets/scss/02-molecules/_action-seq-list.scss
gender404: only reason is copy and paste without noticing

PR #123, opened 2016-10-11
Add self-hosted texta font.

isaacchansky:
- Using the font-stack: `"Texta", "Century Gothic", "Helvetica", "sans-serif"`.
- Self hosting fonts w/ @font-face

**Note:** This does **not** address adjustments for leading, tracking, sizing, etc for the new typeface. Additionally, some items based on `em`s will have to be adjusted, like radio buttons:

screen shot 2016-10-11 at 10 12 59 am

@legostud, re: self-hosting & how I've set up font-face declarations,  let me know what you think of this approach


styleguide/source/assets/scss/06-theme/00-base/_fonts.scss
```diff
@@ -8,115 +8,115 @@
...
+  url("../fonts/texta-blackitalic-webfont.ttf") format("truetype"), url("../fonts/texta-blackitalic-webfont.svg#open_sansregular") format("svg");
   font-weight: 800;
   font-style: italic;
 }
 
 @mixin ma-font($weight) {
   font-weight: $weight;
-  font-family: "Texta", "Century Gothic", "Helvetica", "sans-serif";
+  font-family: "Texta", "Century Gothic", "Helvetica", "Arial", "sans-serif";
 }
```
gender404, line 121: now that we are only using one font-family this mixin is no longer needed.  Also Arial isn't really need either since PC browser automatically use Arial when they see Helvetica.  Doesn't hurt to leave it though.

PR #129, opened 2016-10-12
Design QA updates

reenybeeny:
More updating font sizes, line-heights, etc, for Texta font. Still in progress...

### DIFF
--- a/styleguide/source/assets/scss/02-molecules/_login-util.scss
+++ b/styleguide/source/assets/scss/02-molecules/_login-util.scss
@@ -16,9 +16,10 @@ $bp-login-util-reduce-links-max: "max-width:1120px";
 
     @media ($bp-medium-min) {
       display: table-cell;
+      line-height: 1.3;
       margin-bottom: 0;
       vertical-align: middle;
-      width: 300px;
+      width: 250px;
     }
 
     @media ($bp-login-util-medium-min) {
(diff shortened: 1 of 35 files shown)


styleguide/source/assets/scss/02-molecules/_login-util.scss
```diff
@@ -79,6 +80,14 @@ $bp-login-util-reduce-links-max: "max-width:1120px";
...
 
+      @media ($bp-large-min) {
+        display: inline-block;
+        margin-left: 15px;
+        margin-bottom: 0;
+        padding: 10px;
+        padding-right: 2.5em;
+      }
+
       &:hover {
```
gender404, line 83: you don't need to duplicate the display: inline-block, margin-left: 15px, and margin-bottom: 0;  It should inherit all of those properties from the previous media query

gender404: This doesn't actually do anything.  It's all there is the base and previous breakpoint

 @media ($bp-large-min) {
-        display: inline-block;
-        margin-left: 15px;
-        margin-bottom: 0;
-        padding: 10px;
-        padding-right: 2.5em;
-      }

reenybeeny: updated this

PR #130, opened 2016-10-12
Location Page Header (was location icons)

gender404:
1. New location icons molecule
2. New Location header (imports page header and location icons)
3. four new icons used on the location header
4. renamed folders in atoms so I could add another icon folder


styleguide/tools/gulp/tasks/svg-loc-icons.js
```diff
@@ -0,0 +1,46 @@
+var gulp           = require("gulp"),
```
jasonstanbery, line 1: I _feel_ like this doesn't have to be separate from the other task, but off the top of my head I can't think of how to do it with the normal svg-sprite task. So it's fine as is.

jasonstanbery (approved):

styleguide/tools/gulp/tasks/svg-loc-icons.js
gender404: I was trying to think of a looping method or array variables, but I decided to go with the simpler approach so I could get this done today.

PR #131, opened 2016-10-13
Trying to improve semantic markup where we can.

isaacchansky:
Add <main> tag for core content, <aside> for the sidebar, <nav> where appropriate, use <ol> for breadcrumbs.

### DIFF
--- a/styleguide/source/_patterns/04-templates/single-column.twig
+++ b/styleguide/source/_patterns/04-templates/single-column.twig
@@ -5,11 +5,11 @@
   {% endblock %}
 </div>
 <div class="main-content">
-  <div class="page-content">
+  <main class="page-content">
     {% block pageContent %}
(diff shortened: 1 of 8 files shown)


styleguide/source/_patterns/04-templates/single-column.twig
```diff
@@ -5,16 +5,16 @@
...
 </div>
 <div class="main-content">
-  <div class="page-content">
+  <main class="page-content">
     {% block pageContent %}
       {% include "base-placeholder" with {'placeholder':{'text':'Page content'}} %}
     {% endblock %}
-  </div>
+  </main>
 </div>
```
gender404, line 13: main tag isn't supported in IE.  Edge has support along with other modern browsers.  Just make sure we have shiv for this in place.

gender404: looks like Modernizr supports the main tag

PR #137, opened 2016-10-14
MGVDU-214

jasonstanbery:
Activities component.


gender404:
using :last-child{ border:none} isn't as creative as just using & + & { border-top: 1px solid; }, but It's fine

PR #138, opened 2016-10-14
Location page

gender404:
1. New Action Gallery component
2. changed large breakpoint to 910 instead of 900 to make Action page stack a bit sooner
3. reduced the height of the Action Map on modile


styleguide/source/assets/scss/00-base/_breakpoints.scss
```diff
@@ -3,7 +3,7 @@
 $bp-xs: 480px;
 $bp-s: 620px;  // used in JS files. make sure to update in both places
 $bp-m: 780px;  // used in JS files. make sure to update in both places
-$bp-l: 900px;
+$bp-l: 910px;
```
jasonstanbery, line 6: This makes me a little nervous because I think I used this for the action page activities, and the extra 10 pixels might impact it. But we can go back and add another component level variable if we need to.

jasonstanbery (approved):

PR #141, opened 2016-10-17
Mgvdu 212

jasonstanbery:


styleguide/source/_patterns/05-pages/LOC-Mt-Greylock-State-Park.twig
```diff
@@ -16,4 +16,5 @@
 {% endblock %}
 {% block postContent %}
   {% include "organisms-breadcrumbs" %}
+  {% include "organisms-suggested-pages" %}
```
gender404, line 18: need to reverse the order here.  Breadcrumbs should be last.

styleguide/source/assets/scss/03-organisms/_suggested-pages.scss
```diff
@@ -0,0 +1,61 @@
...
+    }
+  }
+
+  &__items {
+    @include clearfix;
+    margin-bottom: 35px;
+  }
+
+  &__item {
+    display: inline-block;
```
gender404: wouldn't the default display: block be better?

styleguide/source/assets/scss/03-organisms/_suggested-pages.scss
```diff
@@ -0,0 +1,61 @@
...
+    margin: 0;
+    margin-bottom: 30px;
+    vertical-align: top;
+    width: 100%;
+    a {
```
gender404, line 36: missing line break above the a {

styleguide/source/assets/scss/03-organisms/_suggested-pages.scss
```diff
@@ -0,0 +1,61 @@
...
+      height: 100%;
+      width: 100%;
+    }
```
gender404, line 39: I think display: block does this better.  You will probably want font-size: 0 as well to get rid of the slight gap below the image.

styleguide/source/assets/scss/03-organisms/_suggested-pages.scss
```diff
@@ -0,0 +1,61 @@
...
+
+    .ma__decorative-link {
+      display: block;
+    }
+
+    &:first-of-type {
+      border: none;
+    }
+
```
gender404: This feels like it is in the wrong place

styleguide/source/assets/scss/03-organisms/_suggested-pages.scss
```diff
@@ -0,0 +1,61 @@
...
+      float: left;
+      padding: 0 10px;
+      width: 50%;
+    }
+
+    @media ($bp-large-min) {
+      border-left: 1px solid;
+      padding: 0 45px;
+      width: 25%;
+    }
```
gender404, line 57: wrap the border-left in & + & {  }, that will apply it all, but the first item.

styleguide/source/assets/scss/06-theme/03-organisms/_suggested-pages.scss
```diff
@@ -0,0 +1,16 @@
+.ma__suggested-pages {
+  background-color: $c-bg-section;
+
+  &__container {
+    border-color: $c-bd-divider;
+  }
+
+  &__item {
+    border-color: $c-bd-divider;
```
gender404: ideally this should be in the @media, but works as is.  With the suggestion I made above you would need to change it to &__item + &__item { ... } so only the items after the first one get the border

PR #150, opened 2016-10-18
Small form tweaks.

isaacchansky:
- Fix typo of "feeback" => "feedback".
- Scope form button styling to include input[type=button]

### DIFF
--- a/styleguide/source/assets/scss/06-theme/03-organisms/_feedback-form.scss
+++ b/styleguide/source/assets/scss/06-theme/03-organisms/_feedback-form.scss
@@ -28,7 +28,8 @@
   &__controls {
 
     button,
-    input[type="button"] {
+    input[type="button"],
+    input[type="submit"] {
       @include ma-button("normal","solid",$c-theme-primary-1);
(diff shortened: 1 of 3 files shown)


styleguide/source/assets/scss/06-theme/03-organisms/_feedback-form.scss
```diff
@@ -27,12 +27,13 @@
 
   &__controls {
 
-    button {
+    button,
+    input[type="button"] {
       @include ma-button("normal","solid",$c-theme-primary-1);
```
gender404, line 33: is input[type="button"] being used here?  I don't see that in the twig file.

Also in this scenario, the corresponding input would be type="submit" not "button"

isaacchansky: Right, good to have both. The backstory here (apologies for not including that in the description initially) is that we're using the FormStack service to implement the feedback form in the Drupal site which has it's own un-changeable markup it renders.

PR #152, opened 2016-10-19
MGRP-99: Change the search button on keyword search to input.

gleroux02:
Requesting changes for: https://jira.state.ma.us/browse/MGRP-99.


gleroux02:
Sorry I didn't get a chance to write why these changes are needed as this was still a WIP.  I was following the repo protocol of the drupal mass repo and didn't realize that it was a different workflow over here. I'll go ahead and remove this branch and make my changes in a fork as requested.

The reason why we need to switch from a button to an input has to do with how drupal's core js submits its ajax requests. The ajax is applied on the following elements only: input[type=submit] and input[type=image]. So using a button would mean that ajax would not be able to be applied and the page would reload instead. And changing the markup to be an input was the path of least resistance to solve the problem.

Let me close this pr, and remove the branch and get something up over to you in the correct workflow.

PR #153, opened 2016-10-19
MGRP-99: Change search button on keyword search to input to appease drupal ajax.

gleroux02:
Drupal core js will not apply ajax to elements that are not input[type=submit] and input[type=image]. This PR changes the keyword-search button to an input element so that the ajax can properly apply. The js is also altered so that the ".is-dirty" class is applied to the ".ma__action-finder__container" element instead of the form. This is required since the drupal ajax replaces the form on submit and the class would not be retained.

This is required for https://jira.state.ma.us/browse/MGRP-99


gender404:
Ok, we just need to be careful adding Drupal specific changes to the Pattern Library.  The Pattern Library is supposed to be platform agnostic so the components could be used anywhere.  The example we were given was a static html micro site.

I doubt this component would work on a static html site so I think it's fine to convert the button.

Is this pull request ready to be merged?

FYI:  @jasonstanbery

gleroux02:
@legostud understandable. I think that some alterations are inevitable when it comes to form elements. This is not ready quite yet. I'll set this to WIP for now and assign to myself. Should I assign to you or someone else to review once it is ready.

lukewertz:
Closing this PR in favor of a fix within the Drupal implementation.

PR #168, opened 2016-10-31
Use twig namespaces

nstriedinger:
Refractor styleguide for other systems.


lukewertz:
@legostud, Nate put up this PR that should provide the refactor we need to leverage these assets directly in Drupal.

Would you have time today to review this? I'd be happy to hop on a call with you and look at it together.

styleguide/source/_patterns/01-atoms/01-buttons/button-search.twig
```diff
@@ -1,6 +1,6 @@
 <button type="submit" class="ma__button-search">
   <span>Search</span>
   <svg role="img" class="ma__button-search__icon">
-    <use xlink:href="/assets/images/svg-sprite.svg#search"></use>
+    <use xlink:href="/{{ directory }}/images/svg-sprite.svg#search"></use>
   </svg>
```
gender404, line 5: should be {{directory}} not {{ directory }}

styleguide/source/_data/data.json
```diff
@@ -2,6 +2,7 @@
   "title" : "Mass Gov",
   "htmlClass": "pl",
   "bodyClass": "",
+  "directory": "assets",
```
gender404, line 6: I know this is a lot to change, but can we use "assetsDir" instead of "directory" to make it more readable

gender404 (commented):
1. Is this a WIP request or is it ready to be merged (after review changes)?
2. If you open "show pattern info" (gear menu) does the panel still show patterns it contains and where this pattern is included?
3. Have you searched for {% include path with data %} versions (mainly for Rich Text)?

styleguide/source/_patterns/01-atoms/01-buttons/button-search.twig
nstriedinger: Our linters use {{ directory }}, but I can change this and this corroborates http://twig.sensiolabs.org/doc/coding_standards.html

gender404: hmm, if {{ directory }} is the best practice then we can leave it that way, but you'll need to update all of the other files instead which have {{directory}}.  I'll start using that best practice as well on files as I build them.

styleguide/source/_data/data.json
nstriedinger: This just makes it easier for us and builds on our system theme paths. I would have to refractor system paths (Drupal currently), if we change this. If this works and is somewhat arbitrary, lets keep this for the current said system.

styleguide/source/_patterns/01-atoms/01-buttons/button-search.twig
nstriedinger: 1. Ready to be merged
2. This looks to be right.
3. Yes, but this was a massive refractor. I think things are in place.
4. Please review.

gender404:
Looks like we've disabled one of Pattern Labs features:

image

nstriedinger:
I think this is a bug in PatternLabs. I've reported the issue.

PR #169, opened 2016-10-31
Design QA - contact us section

reenybeeny:
Updated line-height and padding on address in contact us section


styleguide/source/assets/scss/03-organisms/_page-header.scss
```diff
@@ -37,4 +37,13 @@
...
   }
+
+ &__intro {
+
+   p {
+     @media ($bp-small-min) {
+       font-size: 1.625rem;
+     }
+   }
+ }
```
gender404: Just double checking - Are you okay with this font change appearing on other pages (L2, Action, Location, General)?

PR #182, opened 2016-11-08
callout time design QA

reenybeeny:
Changed font-weight, size, and style for callout time. Also update margins so there’s more space below the icon on mobile.

### DIFF
--- a/styleguide/source/assets/scss/03-organisms/_callout-time.scss
+++ b/styleguide/source/assets/scss/03-organisms/_callout-time.scss
@@ -10,7 +10,8 @@
   &__icon {
     font-size: 0;
     height: 30px;
-    margin: 12px 12px 12px 0;
+    margin: 12px;
+    margin-left: 0;
     width: 30px;
   }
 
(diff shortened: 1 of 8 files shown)


styleguide/source/assets/scss/03-organisms/_callout-time.scss
```diff
@@ -10,11 +10,13 @@
   &__icon {
     font-size: 0;
     height: 30px;
-    margin-right: 12px;
+    margin: 12px 12px 12px 0;
     width: 30px;
```
gender404, line 15: I usually write these as margin: 12px;  margin-left: 0; but this is fine.

styleguide/source/assets/scss/03-organisms/_callout.scss
```diff
@@ -3,33 +3,39 @@
   padding-left:38px;
   margin-bottom: 45px;
 
+  @media ($bp-x-small-max) {
+    padding-left:20px;
+  }
+
+
```
gender404, line 10: can you reverse this to do mobile first instead.  $bp-x-small-min { padding-left: 38px; }  It's just a bit more consistent

PR #189, opened 2016-11-10
312 contact group - refactor address

gender404:
Requested Change by Palentir to refactor the Contact Group molecule to work for an address entered via a textarea field.

1. Contact Group - added an address type section of code
2. Contact Us - removed the address specific section and replaced it with a contact group
3. Content Link - Removed "target" and added a "--phone" modifier to the styling
4. Went through all json files to update the contactUs formating to match the new code


styleguide/source/_patterns/02-molecules/contact-us.twig
```diff
@@ -13,56 +11,13 @@
     </h5>
   {% endif %}
 
-  {% if contactUs.phone %}
-    {% set contactGroup = contactUs.phone %}
+  {% for contactGroup in contactUs.groups %}
+    {% if contactGroup.hidden and not accordion %}
```
jesconstantine, line 15: @legostud : In MQA-64 (https://jira.state.ma.us/browse/MQA-64) @reenybeeny writes that the first two contact groups should be visible before applying this accordion functionality.  Is this work based on that spec?  (I suspect not since we check for a passed value of contactGroup.hidden)  Could we instead keep track of the index of this for loop and assign a value to the accordion variable accordingly rather than have it be sent as an implementation variable?

gender404: yeah, this needs to be refactored.  I was thinking of doing two for loops.  First loop through the first two groups.  Second loop through what's left inside an accordion content div.

jesconstantine: Sounds good - I'll give the refactor a shot and link to PR here.

PR #214, opened 2016-11-28
Change <h3> in header banner to be a <p> w/ class.

isaacchansky:
- Change `<h3>` in header banner to be a `<p>` w/ class.
- Make headline sizing mixin-based for re-use, to decouple semantics from styles

@legostud I'm not sure I like how I've accomplished this, entirely. 
I went back & forth between using placeholders & mixins, and went w/ mixins because of placeholders limitations in media queries & I don't *think* we can get around that. 
Let me know if what I've done jives with how you imagined this working & if you're into it, if not I'd love some feedback!

### DIFF
--- a/styleguide/source/_patterns/03-organisms/by-template/collage-header.twig
+++ b/styleguide/source/_patterns/03-organisms/by-template/collage-header.twig
@@ -35,10 +35,10 @@
         <h1>Welcome to the Mass.Gov Pilot</h1>
       </div>
       <div class="ma__collage-header__details">
-        <p class="ma__collage-header__details__lead">We're redesigning this website to make it easier to find the things you need.</p>
-        <p class="ma__collage-header__details__content">This is a big project, so we’re going public bit by bit, right here where you can see it and be a part of it. We hope you’ll explore the new site and provide some feedback so we can keep improving it as we add more information and features.</p>
+        <p class="ma__collage-header__lead">We're redesigning this website to make it easier to find the things you need.</p>
+        <p class="ma__collage-header__description">This is a big project, so we’re going public bit by bit, right here where you can see it and be a part of it. We hope you’ll explore the new site and provide some feedback so we can keep improving it as we add more information and features.</p>
 
-        <p class="ma__collage-header__details__content">Thank you for being a part of this project!</p>
+        <p class="ma__collage-header__description">Thank you for being a part of this project!</p>
       </div>
     </div>
   </div>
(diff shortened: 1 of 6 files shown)


gender404:
I like the mixins.  They make a lot of sense and help with consistency across the site.  I'm not sure converting the h3 to a p tag is the best approach though.  If we want to go down this route then I would suggest adding a class to the existing p tag as well and apply styling to each without the overrides for "lead".

Another option would be to make the h3 a div div instead.

Also keep in mind that this header is currently just for the Pilot homepage and isn't slated to exist after the new home page is created.

styleguide/source/assets/scss/03-organisms/_collage-header.scss
```diff
@@ -137,13 +137,9 @@ $collage-header-bp-min: "min-width: 701px";
...
-      h3 {
-        line-height: 1.3;
-      }
-
-      p {
-        font-size: 1.625rem;
+      font-size: 1.625rem;
+      &__lead {
+        @include ma-h3;
       }
```
gender404: quick note here.  Attributes should be in alphabetical order and there should be a line break above &__lead.

isaacchansky:
Cool, went with the more deliberate classes on the <p> tags rather than overrides. Thanks!

styleguide/source/_patterns/03-organisms/by-template/collage-header.twig
```diff
@@ -36,9 +36,9 @@
       </div>
       <div class="ma__collage-header__details">
         <p class="ma__collage-header__details__lead">We're redesigning this website to make it easier to find the things you need.</p>
-        <p>This is a big project, so we’re going public bit by bit, right here where you can see it and be a part of it. We hope you’ll explore the new site and provide some feedback so we can keep improving it as we add more information and features.</p>
+        <p class="ma__collage-header__details__content">This is a big project, so we’re going public bit by bit, right here where you can see it and be a part of it. We hope you’ll explore the new site and provide some feedback so we can keep improving it as we add more information and features.</p>
```
gender404, line 40: grandchildren aren't allowed in BEM.  If you run into needing grandchilden it's usually a sign that you need to split out that section into it's own molecule.  In this case I think you can just use ma__collage-header__lead and ma__collage-header__description

isaacchansky: Right, updated. That feels a lot better, good catch.

styleguide/source/assets/scss/06-theme/03-organisms/_collage-header.scss
```diff
@@ -20,13 +20,10 @@ $collage-header-bp-max: "max-width: 700px";
   }
 
   &__details {
+    font-weight: 300;
+  }
```
gender404, line 25: Sorry to be a pain, but can you change this to &__description.

isaacchansky: ah sorry about that, udpated

PR #219, opened 2016-11-29
Add pilot-specific js module to handle the following:

isaacchansky:
- keep top-level-links that have sub-menus inert when clicked
- clear out open submenus when mobile sidebar is closed
- anchor links that are top-level-links close sidebar when clicked

### DIFF
--- a/styleguide/source/_patterns/02-molecules/main-nav.twig
+++ b/styleguide/source/_patterns/02-molecules/main-nav.twig
@@ -2,7 +2,7 @@
   <ul class="ma__main-nav__items js-main-nav">
     {% for nav in mainNav %}
       <li class="ma__main-nav__item {{ nav.active ? 'is-active' : '' }} {{ nav.subNav ? 'has-subnav js-main-nav-toggle' : '' }}">
-        <a class="ma__main-nav__top-link" href="{{nav.href}}" title="{{nav.text}}">{{nav.text}}</a>
+        <a class="ma__main-nav__top-link js-main-nav-top-link" href="{{nav.href}}" title="{{nav.text}}">{{nav.text}}</a>
         {% if nav.subNav %}
           <div class="ma__main-nav__subitems js-main-nav-content is-closed">
             <ul class="ma__main-nav__container">
--- a/styleguide/source/_patterns/02-molecules/main-nav.twig
+++ b/styleguide/source/_patterns/02-molecules/main-nav.twig
@@ -1,8 +1,8 @@
 <section class="ma__main-nav">
   <ul class="ma__main-nav__items js-main-nav">
     {% for nav in mainNav %}
-      <li class="ma__main-nav__item {{ nav.active ? 'is-active' : '' }} {{ nav.subNav ? 'has-subnav js-main-nav-toggle' : '' }}">
-        <a class="ma__main-nav__top-link js-main-nav-top-link" href="{{nav.href}}" title="{{nav.text}}">{{nav.text}}</a>
+      <li class="ma__main-nav__item {{ nav.active ? 'is-active' : '' }} {{ nav.subNav ? 'has-subnav js-main-nav-toggle' : 'js-main-nav-top-link' }}">
+        <a class="ma__main-nav__top-link" href="{{nav.href}}" title="{{nav.text}}">{{nav.text}}</a>
         {% if nav.subNav %}
           <div class="ma__main-nav__subitems js-main-nav-content is-closed">
             <ul class="ma__main-nav__container">
--- a/styleguide/source/_patterns/06-pilot/main-nav.twig
+++ b/styleguide/source/_patterns/06-pilot/main-nav.twig
@@ -1,8 +1,8 @@
 <section class="ma__main-nav">
   <ul class="ma__main-nav__items js-main-nav">
     {% for nav in mainNavPilot %}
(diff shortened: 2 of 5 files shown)


styleguide/source/_patterns/02-molecules/main-nav.twig
```diff
@@ -2,7 +2,7 @@
   <ul class="ma__main-nav__items js-main-nav">
     {% for nav in mainNav %}
       <li class="ma__main-nav__item {{ nav.active ? 'is-active' : '' }} {{ nav.subNav ? 'has-subnav js-main-nav-toggle' : '' }}">
-        <a class="ma__main-nav__top-link" href="{{nav.href}}" title="{{nav.text}}">{{nav.text}}</a>
+        <a class="ma__main-nav__top-link js-main-nav-top-link" href="{{nav.href}}" title="{{nav.text}}">{{nav.text}}</a>
         {% if nav.subNav %}
```
gender404, line 6: I would suggest moving this JS class into the nav.SubNav conditional on the li:
{{ nav.subNav ? 'has-subnav js-main-nav-toggle' : 'js-main-nav-top-link' }}

This will make it easy to right JS that works for either case.

styleguide/source/_patterns/06-pilot/main-nav.twig
```diff
@@ -2,7 +2,7 @@
   <ul class="ma__main-nav__items js-main-nav">
     {% for nav in mainNavPilot %}
       <li class="ma__main-nav__item {{ nav.active ? 'is-active' : '' }} {{ nav.subNav ? 'has-subnav js-main-nav-toggle' : '' }}">
-        <a class="ma__main-nav__top-link" href="{{nav.href}}" title="{{nav.text}}">{{nav.text}}</a>
+        <a class="ma__main-nav__top-link js-main-nav-top-link" href="{{nav.href}}" title="{{nav.text}}">{{nav.text}}</a>
         {% if nav.subNav %}
```
gender404, line 6: I would suggest moving this JS class into the nav.SubNav conditional on the li:
{{ nav.subNav ? 'has-subnav js-main-nav-toggle' : 'js-main-nav-top-link' }}

This will make it easy to right JS that works for either case.

styleguide/source/assets/js/modules/mainNav.js
```diff
@@ -154,7 +154,7 @@ export default function (window,document,$,undefined) {
         hide($openContent);
       }
     });
-    $mainNavToggle .on('click', function(e) {
+    $mainNavToggle.on('click', function(e) {
       if(windowWidth <= breakpoint) {
```
gender404, line 158: oops.  Thanks for catching this.

styleguide/source/assets/js/modules/mainNavPilot.js
```diff
@@ -0,0 +1,57 @@
...
+    $parent.find('.ma__main-nav__item:not(.has-subnav) .js-main-nav-top-link').click(function() {
+      $('body').removeClass('show-menu');
+    });
+
+    // Hide any open submenu content when the sidebar menu is closed
+    $('.js-header-menu-button').click(function() {
+      let $openContent = $parent.find('.js-main-nav-content.' + openClass);
+      hide($openContent);
+    });
+
```
gender404: This seems like a useful event handler beyond Pilot so maybe it should live in the manNav.js instead.

styleguide/source/assets/js/modules/mainNavPilot.js
```diff
@@ -0,0 +1,57 @@
...
+    // make root top-level links inert for pilot
+    $parent.find('.has-subnav .js-main-nav-top-link').click(function(e) {
+      e.preventDefault();
+    });
+
+    // Ensure top-level links that are potential anchor links close the sidebar on mobile
+    $parent.find('.ma__main-nav__item:not(.has-subnav) .js-main-nav-top-link').click(function() {
+      $('body').removeClass('show-menu');
+    });
+
```
gender404, line 16: I think this can be simplified to (assuming twig changes above).  

$parent.find('.js-main-nav-top-link').on('click', function() {
     $('.js-close-sub-nav').trigger('click');
});

styleguide/source/assets/js/modules/mainNavPilot.js
```diff
@@ -0,0 +1,57 @@
...
+            .slideDown(0);
+        });
+      }
+    }
+
+    // make root top-level links inert for pilot
+    $parent.find('.has-subnav .js-main-nav-top-link').click(function(e) {
+      e.preventDefault();
+    });
+
```
gender404, line 11: Based on the Twig changes this would become $mainNavToggle.find('a').on('click', function(e) {

styleguide/source/assets/js/modules/mainNavPilot.js
```diff
@@ -0,0 +1,57 @@
...
+        $content
+        .stop( true, true )
+        .slideUp('fast',function() {
+          $content
+            .addClass(closeClass)
+            .slideDown(0);
+        });
+      }
+    }
+
```
gender404, line 6: This shouldn't be needed if we move the other event handler below to mainNav.js

gender404 (commented):
Let me know if these comments make sense

isaacchansky:
Those all make a lot of sense, really like not having to duplicate any code. 👍 Thanks for the feedback!

PR #227, opened 2016-12-01
Fix a11y keyboard navigation for pilot menu

isaacchansky:
The left/right keys are not behaving correctly. (i.e you can't jump between top-level links)

Since we chanced the non-dropdown classes in another PR, this needed adjustment as well.

### DIFF
--- a/styleguide/source/assets/js/modules/mainNav.js
+++ b/styleguide/source/assets/js/modules/mainNav.js
@@ -11,7 +11,7 @@ export default function (window,document,$,undefined) {
         closeClass = "is-closed",
         submenuClass = "show-submenu",
         $parent = $(this),
-        $mainNavToggle = $parent.find('.js-main-nav-toggle'),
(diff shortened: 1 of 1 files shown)


styleguide/source/assets/js/modules/mainNav.js
```diff
@@ -11,7 +11,7 @@ export default function (window,document,$,undefined) {
         closeClass = "is-closed",
         submenuClass = "show-submenu",
         $parent = $(this),
-        $mainNavToggle = $parent.find('.js-main-nav-toggle'),
+        $mainNavToggle = $parent.find('.js-main-nav-toggle, .js-main-nav-top-link'),
         previousKey = null,
```
gender404, line 15: not sure about doing this.  There are various events that are tied to this element related to showing and hiding the drop down menus, but I can't think of a better solution.

gender404 (approved):
we probably should refactor this component to make the code a bit cleaner in future now that we have more knowledge about how it's supposed to function.

PR #239, opened 2016-12-07
244 emergency alerts

gender404:
New Amergency Alerts Organism
https://jira.state.ma.us/browse/MGRP-278

1. New Alert Button, Emergency Header, and Emergency Alert molecules created for this to consume.
2. New Emergency Alerts JS file that records what state (open/closed) the accordion should be in on page load.
3. Cookies JS - modified the setCookie function to make 'expire' and optional argument that when left blank results in a cookie that expires when the session expires.
4. Added a "no-js" css class to the html element (This will need to be updated by Palantir @lukewertz)

Notes:
1. This Organism was added to the top of all of the templates except the narrow-template.
2. Data for this Organism was added the three Guide pages for testing purposes.
3. see md file for additional notes

### DIFF
--- a/styleguide/source/_patterns/02-molecules/emergency-alert.md
+++ b/styleguide/source/_patterns/02-molecules/emergency-alert.md
@@ -10,10 +10,10 @@ Link is optional.
 messages are sorted with the newest one first
 
 ### Used in: 
-organisms/by-template/emergency-alerts.twig 
+[@organisms/by-template/emergency-alerts.twig](/?p=organisms-emergency-alerts)
 
 ### Contains
-atoms/11-text/link.twig
+[@atoms/11-text/link.twig](/?p=atoms-link)
 
 ### Variables 
 ~~~ 
--- a/styleguide/source/_patterns/02-molecules/emergency-alert.md
+++ b/styleguide/source/_patterns/02-molecules/emergency-alert.md
@@ -1,0 +1,35 @@
+---
+title: Emergency Alert
+---
+Description: Emergency Alert message describing the event with a timestamp reflecting when this message was created.  An optional link is provided to take the user to page with more information
+
+## State: ALPHA
+
+###  Notes
+Link is optional.
+messages are sorted with the newest one first
+
+### Used in: 
+organisms/by-template/emergency-alerts.twig 
+
+### Contains
+atoms/11-text/link.twig
+
+### Variables 
+~~~ 
+emergencyAlert: {
+  message: 
+    type: string/required,
+  timeStamp: 
+    type: date/required
+    format: mm.dd.yy, h:mm am/pm
+  link: {
+    href: 
+      type: string/url,
+    text:
+      type: string,
+    chevron:
+      type: string => always "true"
+  }
+}
+~~~
--- a/styleguide/source/_patterns/02-molecules/emergency-alert.twig
+++ b/styleguide/source/_patterns/02-molecules/emergency-alert.twig
@@ -1,0 +1,7 @@
+<div class="ma__emergency-alert">
+  <div class="ma__emergency-alert__message">{{ emergencyAlert.message }} <span class="ma__emergency-alert__time-stamp">{{ emergencyAlert.timeStamp }}</span></div>
+  <div class="ma__emergency-alert__link">
+    {% set link = emergencyAlert.link %}
+    {% include "@atoms/11-text/link.twig" %}
+  </div>
+</div>
--- a/styleguide/source/_patterns/02-molecules/emergency-alert.twig
+++ b/styleguide/source/_patterns/02-molecules/emergency-alert.twig
@@ -1,7 +1,9 @@
 <div class="ma__emergency-alert">
   <div class="ma__emergency-alert__message">{{ emergencyAlert.message }} <span class="ma__emergency-alert__time-stamp">{{ emergencyAlert.timeStamp }}</span></div>
   <div class="ma__emergency-alert__link">
-    {% set link = emergencyAlert.link %}
-    {% include "@atoms/11-text/link.twig" %}
+    {% if emergencyAlert.link %}
+      {% set link = emergencyAlert.link %}
+      {% include "@atoms/11-text/link.twig" %}
+    {% endif %}
   </div>
 </div>
--- a/styleguide/source/_patterns/02-molecules/emergency-header.twig
+++ b/styleguide/source/_patterns/02-molecules/emergency-header.twig
@@ -1,7 +1,7 @@
-<section class="ma__emergency-header">
+<div class="ma__emergency-header">
   <div class="ma__emergency-header__label">
     {% include "@atoms/05-icons/svg-alert.twig" %}
     <span>Emergency Alerts</span>
   </div>
-  <h1 class="ma__emergency-header__title">{{ emergencyHeader.title }}</h1>
-</section>
+  <h2 class="ma__emergency-header__title">{{ emergencyHeader.title }}</h2>
+</div>
--- a/styleguide/source/_patterns/02-molecules/emergency-header.twig
+++ b/styleguide/source/_patterns/02-molecules/emergency-header.twig
@@ -1,0 +1,7 @@
+<section class="ma__emergency-header">
+  <div class="ma__emergency-header__label">
+    {% include "@atoms/05-icons/svg-alert.twig" %}
+    <span>Emergency Alerts</span>
+  </div>
+  <h1 class="ma__emergency-header__title">{{ emergencyHeader.title }}</h1>
+</section>
--- a/styleguide/source/_patterns/03-organisms/by-template/emergency-alerts.json
+++ b/styleguide/source/_patterns/03-organisms/by-template/emergency-alerts.json
@@ -1,5 +1,6 @@
 {
   "emergencyAlerts": {
+    "id":"GUID18378923w38789",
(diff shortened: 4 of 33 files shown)


gender404:
This is ready to be merged

styleguide/source/_patterns/02-molecules/emergency-header.twig
```diff
@@ -0,0 +1,7 @@
+<section class="ma__emergency-header">
+  <div class="ma__emergency-header__label">
+    {% include "@atoms/05-icons/svg-alert.twig" %}
+    <span>Emergency Alerts</span>
+  </div>
+  <h1 class="ma__emergency-header__title">{{ emergencyHeader.title }}</h1>
```
jesconstantine: I wonder if this should be an H2?  Should there only be 1 H1 per page (the page title)?

styleguide/source/_patterns/03-organisms/by-template/emergency-alerts.md
```diff
@@ -0,0 +1,57 @@
...
+  emergencyHeader: {
+    title: 
+      type: string/required
+  },
+
+  alerts: [{
+    message: 
+      type: string/required
+    timeStamp: 
+      type: string/required
```
jesconstantine: Can we dictate a spec for the timestamp?  Like a template of mm.dd.yy hh:mm am/pm

jesconstantine (commented):
I just had a couple questions, but not necessarily requesting any changes.  Thanks so much for tagging me here as a reviewer.  It will be helpful next sprint when we implement in Drupal!

styleguide/source/_patterns/02-molecules/emergency-header.twig
gender404: @isaacchansky - Can the first header on the page be an H2 instead of an H1?

Normally, I would use an h2 as the default header tag in a section, but with the accessibility changes, I'm not sure it that is correct.

styleguide/source/_patterns/03-organisms/by-template/emergency-alerts.md
gender404: that's not a bad idea

styleguide/source/_patterns/02-molecules/emergency-header.twig
jesconstantine: This is tricky.  I do believe that the only `<h1>` on the page should be reserved for the page title.  But I also understand and agree that there shouldn't be any other headings before that semantically ( https://dequeuniversity.com/assets/html/jquery-summit/html5/slides/headings.html ) -- although that is what we do on classic massgov for the alert banner and main navigation.

Not entirely sure what the best way forward is here.  @ygannett do you have any thoughts on the matter?

styleguide/source/_patterns/03-organisms/by-template/emergency-alerts.md
jesconstantine: Thanks!

styleguide/source/_patterns/02-molecules/emergency-alert.twig
```diff
@@ -0,0 +1,7 @@
+<div class="ma__emergency-alert">
+  <div class="ma__emergency-alert__message">{{ emergencyAlert.message }} <span class="ma__emergency-alert__time-stamp">{{ emergencyAlert.timeStamp }}</span></div>
+  <div class="ma__emergency-alert__link">
```
gleroux02, line 3: @legostud If the link is optional then there should be an if statement around the link markup so that it will only display if the link is not empty.

styleguide/source/_patterns/02-molecules/emergency-alert.md
```diff
@@ -0,0 +1,35 @@
...
+Description: Emergency Alert message describing the event with a timestamp reflecting when this message was created.  An optional link is provided to take the user to page with more information
+
+## State: ALPHA
+
+###  Notes
+Link is optional.
+messages are sorted with the newest one first
+
+### Used in: 
+organisms/by-template/emergency-alerts.twig
```
gleroux02: Let's make this a link to the organism. @jesconstantine started to do this with her work on atoms and I think it would be good to carry through. 

```
- [@organisms/by-template/emergency-alerts](/?p=organisms-emergency-alerts)
```

styleguide/source/_patterns/02-molecules/emergency-alert.md
```diff
@@ -0,0 +1,35 @@
...
+
+###  Notes
+Link is optional.
+messages are sorted with the newest one first
+
+### Used in: 
+organisms/by-template/emergency-alerts.twig 
+
+### Contains
+atoms/11-text/link.twig
```
gleroux02: Same as above let's make this a link. To avoid writing this elsewhere can you look through all of the documentation you've added and add links where it makes sense?

gleroux02: With HTML5 I do believe it is safe to use multiple H1s on a page, but they need to be contained within an article, section, nav or aside. This is within a section tag so it should be representing the h1 within the given section and not the main content of the page. I know that the W3c suggests using article instead of section when:

> the content is complete, or self-contained, composition.

Would that be the case in this situation?

gleroux02: Actually now that I think about this some more. Why is this in a section tag at all? If the purpose of this molecule is for it to be incorporated into an organism that is it's own section of the page why would we need an additional section here? If this is only used for a style to be applied to it then it should be a div and not a section tag.

I was trying to figure out if this needed to be a molecule at all or should it be an atom with the wrapping markup provided by the organism instead. But I'm on the fence here. I think I'd just prefer if the json didn't require a separate object for the title of the emergencyAlert.

styleguide/source/_patterns/03-organisms/by-template/emergency-alerts.json
```diff
@@ -0,0 +1,38 @@
+{
+  "emergencyAlerts": {
+    "id":"GUID18378923w38789",
+    "buttonAlert": {
+      "hideText": "Hide",
+      "showText": "Show",
+      "text": "Updates"
+    },
+    "emergencyHeader": {
```
gleroux02, line 9: Why does this require a separate object for the title of the emergencyAlert section?

styleguide/source/_patterns/03-organisms/by-template/emergency-alerts.json
```diff
@@ -0,0 +1,38 @@
+{
```
gleroux02, line 1: The setup for this suggests that there would only be one alert with multiple informational updates that relate to it. However the name is emergencyAlerts plural. Is there a scenario where a separate alert for a different situation needs to be displayed along with an existing unrelated alert? I don't know if this was a requirement from the ticket but it is something that popped into my head when reviewing this. The current setup wouldn't provide a solution for this scenario.

gleroux02 (commented):
@legostud I have a few open questions I'd like you to take a look at.

styleguide/source/_patterns/02-molecules/emergency-alert.md
gender404: 1. I can add links to the md files
2. h1 still feels weird.  It is part of a section tag, but I'm going to change it to an h2 instead just because I still feel pages should only have one h1 on them (page title)
3. I replaced the section tag on the header with a div

I made emergency alert a molecule, because it is a chunk of related content (icon, label, title).  It could have been a molecule (one field), but it didn't feel right putting it there.  I think the single html element rule should still apply for atoms.  atoms would then be considered a single html element with a single input field.

styleguide/source/_patterns/03-organisms/by-template/emergency-alerts.json
gender404: The requirement was to have a single Alert Organism per page.  The title would just be adjusted to contain both events.

It's plural because it could be for more than one alert.  Winter Storm and Boston Marathon alerts.

The individual emergency alert molecule now feels more like an alert detail though.  Not sure it's important enough to refactor that though since it could affect naming issues when I build the Alert Detail page.

styleguide/source/_patterns/03-organisms/by-template/emergency-alerts.json
gender404: each molecule, atom, organism gets it's own json object.  I could refactor this to be emergencyAlerts.title, but then I'd need to convert it in the twig file to emergencyHeader.title so it can be passed properly to the emergency header twig file.  This seemed to be the simpler approach

styleguide/source/_patterns/03-organisms/by-template/emergency-alerts.json
gleroux02: Can we show an example of this or add it to the notes? As the direction isn't clear otherwise?

styleguide/source/_patterns/03-organisms/by-template/emergency-alerts.json
gleroux02: I think we've stumbled on my biggest complaint about patternlab as far as trying to implement it within a different system. It's cumbersome to create the variable structure needed when not using json. It's simpler when implementing to setup an object of: emergencyAlert -> title, id, button -> button variables, alerts -> alert item. In this case it is a small example and not overly burdensome and I lean more towards letting it in.

PR #238, opened 2016-12-06
MGRP-258 document mayflower atoms

jesconstantine:
This PR adds documentation to Mayflower atoms with the exception of icon pattern variations (see discussion below).   

Addresses ticket: https://jira.state.ma.us/browse/MGRP-258

### Acceptance notes
Each component should have a file named "name-of-template".md where the name of template is the same as the twig and json file.
The document should set the development status for each component of: depreciated, alpha, beta or stable.
The document should follow the example pattern below
Update the readme to include information about documentation.

The documentation is in the form of `.md` files located within the same directory as the pattern which they are documenting.

### Example .md template:

> ---
> title: Button main
> ---
> Description: an atom for a link styled as a button.
> ## State: ALPHA
> ###  Notes
> Describe the roadmap to get the component to a stable state. And any other notes for rendering.
> ### Used in: 
> List of atoms/molecules/organisms used in. 
> ### Contains
> List of atoms/molecules/organisms that are used.
> ### Variables 
> ~~~ 
> buttonMain {
>     href:
>         type: string/required
>     text: 
>         type: string/required
> }
> ~~~

### To Test:
1. Pull the dev branch from this repo (PR has been merged)
2. From within `styleguide/`, run `gulp` to build Mayflower
3. Visit localhost:3000
4. Use the top blue navigation menu to browse through the list of `ATOMS`, note patterns which have DEPRECATED status are indicated in menu (gif below)
5. Click any atom pattern, show the pattern info (under cog icon in top right corner of nav) (see gif below)
6. Bottom half of your window should change to show pattern information with twig/html tabs on the right, and documentation on the left.  This left pane is populated via the `.md` files which live in the codebase right next to (and named the same as) the twig templates which they describe. (see gif below)

[https://gyazo.com/b5fcf0ada7fe1bb6a1a4049b1d61ac42](https://gyazo.com/b5fcf0ada7fe1bb6a1a4049b1d61ac42)
### Notes:

I think it would be good for someone to revisit the development state that I used for each of these patterns.  I labelled most alpha with the exception of maybe one or two that could be deprecated.

PR #240, opened 2016-12-07
Mgrp 258 styleguide documentation

gleroux02:
WIP need to work with @jesconstantine to merge this work with #238


gender404:
Can you guys do one more task while you're working on this documentation?  I'm wondering if any of these atoms or molecules are in the incorrect location in PatternLab.  Maybe we could record a list of items that should be re-categorized into a more logical location.

Luke and I had slightly different definitions for Atoms and Molecules.  I think his, based on number of fields, makes more sense than Brad Frosts based on the number of html elements.  So Atoms should only contain one field and molecules would have multiple fields.

Thoughts?

jesconstantine:
I'm up for that, @legostud .  I would also love it if we could document our consensus and provide a definitive example of each pattern type here, that way we have a map to work from.  And is there a place where that content can / should live in mayflower docs?

gleroux02:
I'd like to bring this conversation into the ticket so that it doesn't get lost after this PR is merged. Also I switched the base of this PR so that it is against @jesconstantine's branch. I'll fix the conflicts as we had both added documentation for buttons.

PR #245, opened 2016-12-08
Sprint8 accessibility improvements

gender404:
1. New optional "info" variable added to the "decorative links" atom 
  * this contains hidden text for screen reads that is added to the link in cases where the link's text is too generic like "learn more"
2. New optional "sub" variable added to "section links" molecule
  * if sub equals "true" then the section link title will be rendered as an h3, otherwise an h2 will be used.
  * this change is "mostly" backward compatible.  You can have an h2 followed by another h2, but it is not semantic.
3. New optional "id" variable added to the "sections links" molecule
  * this value is used to connect the title and nav elements
  * this value has to be unique for each section links molecule used on the page.


gender404:
@ygannett - if this looks okay feel free to merge it into your branch.

PR #249, opened 2016-12-14
Emergency alerts design QA

reenybeeny:
- updated some font sizes, padding, and letter spacing
- added a button hover state

### DIFF
--- a/styleguide/source/assets/scss/02-molecules/_emergency-alert.scss
+++ b/styleguide/source/assets/scss/02-molecules/_emergency-alert.scss
@@ -23,10 +23,4 @@
       }
     }
   }
-
-  &:last-child {
-    @media ($bp-small-max) {
-      padding-bottom: 20px;
-    }
-  }
 }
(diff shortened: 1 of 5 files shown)


styleguide/source/assets/scss/02-molecules/_emergency-alert.scss
```diff
@@ -21,6 +22,11 @@
         border-bottom-width: 3px;
       }
     }
+  }
 
+  &:last-child {
+    @media ($bp-small-max) {
+      padding-bottom: 20px;
+    }
   }
```
gender404: I would add this to ma__emergency-alerts__container or ma__emergency-alerts__content instead of using a last-child, which could break it something besides an alert was added after the last alert.

reenybeeny: Updated. thanks!

reenybeeny: I updated this again based on your feedback to put it directly on the container

PR #251, opened 2016-12-15
CatIcon - using file path in JSON and changing variable from svg to icon

gender404:


gender404:
This is related to https://github.com/massgov/mass/pull/267

GBirch:
Please see comment in https://github.com/massgov/mass/pull/267 -- it seems likely that this will break a certain amount of stuff in Drupal

gender404:
@GBirch - I've updated everything in MayFlower that uses the catIcon object.

PR #256, opened 2016-12-20
MGRP-302 Action Finder

gender404:
https://jira.state.ma.us/browse/MGRP-302

Updates to the Action Finder Organism to make it work on Org Pages.

1. Filter Search input is now included only if there are more than six general results
2. "Featured:" Heading is now content using the "featuredHeading" variable.
3. "All Action & Guides:" is now content using the "generalHeading" variable.
4. New optional See All link added to the bottom of the organism.
5. Small styling tweak to remove "ma__action-finder__banner" html element to simplify the code.


styleguide/source/_patterns/03-organisms/by-author/action-finder.twig
```diff
@@ -1,4 +1,4 @@
-<section class="ma__action-finder">
+<section class="ma__action-finder" id="GUID831741781374893">
```
patrickfweston, line 1: @legostud : I realize I'm new here, but could you explain why we're using a GUID here? Is it just to apply the background image via CSS, rather than trying to do it via an inline style?

gender404: Since the background image switch based on the browser width (for performance) we cannot do a inline background style.  Using a unique id is needed so we can target that specific element with the CSS within the style block.

patrickfweston: Got it. Thank you!

styleguide/source/_patterns/03-organisms/by-author/action-finder.twig
```diff
@@ -10,15 +10,16 @@
       }
     }
   </style>
-  <div id="GUID831741781374893" class="ma__action-finder__banner"></div>
   <div class="ma__action-finder__container">
     <header class="ma__action-finder__header">
       <h2 class="ma__action-finder__title">What Would You Like to Do?</h2>
```
patrickfweston: Should this be a variable and not hard coded as well?

gender404: probably, I'll make that update as well.

PR #257, opened 2016-12-20
MGRP-305 - Sections Three Up - refactored to use comp-title

gender404:


gleroux02 (commented):
Should the title be centered? MGRP-305 says to left align the title. But I could be looking at the wrong title.
image

gender404:
Your screen shot looks odd, because it has two underlines.  Maybe it's the old version.
image

gender404:
The comp heading atom now accepts a centered option so it can be either.  On the subtopic page it is supposed to be centered => http://dev.massgov.velir.com/?p=pages-L1-state-parks-and-recreation

On the org page it should be left aligned.

gleroux02:
Looks like the issue was that the css wasn't being compiled properly. This should be resolved with the change I just pushed.

PR #258, opened 2016-12-21
300 icon list

gender404:
https://jira.state.ma.us/browse/MGRP-300

1. Updated Social Icon SVGs to not included the green background
2. Refactored Social Icons twig file to use JSON data and the new social icons.
3. Added new flickr svg icon
4. New Icon Links organism


styleguide/source/_patterns/01-atoms/05-icons/svg-blog.twig
```diff
@@ -0,0 +1,4 @@
+<svg focusable="false" role="img" class="svg-blog">
```
gleroux02, line 1: The svg twig files need to have a corresponding .md file that has description information. It should also be marked as deprecated as we will be eventually refactoring how the icons are built.

styleguide/source/_patterns/02-molecules/social-links.twig
```diff
@@ -1,52 +1,13 @@
-<section class="ma__social-links">
```
gleroux02, line 1: There needs to be an updated social-links.md file that includes the json variables required to render this molecule.

styleguide/source/_patterns/03-organisms/by-author/icon-links.twig
```diff
@@ -0,0 +1,16 @@
+{% include "@atoms/04-headings/sidebar-heading.twig" %}
```
gleroux02: Please add a corresponding icon-links.md file for this organism. Also should this be a molecule instead of an organism? If it only includes atoms I don't see why it needs to be an organism.

gleroux02 (changes requested):

gleroux02 (approved):

PR #260, opened 2016-12-22
[Do not merge] Proof-of-concept tracking utility classes.

isaacchansky:
Add `.trk-` classes for atoms/molecules/organisms in POC for sidebar targeting.

The goal of this is the establish a convention of `.trk-` prefixed classes to exist by default in atoms/molecules/organisms so we can traverse the DOM in a more fine-grained way to track conversions. 

For this POC I just filled out the sidebar of action pages. 
Check out an action page (http://localhost:3000/patterns/05-pages-ACTION-get-a-state-park-pass/05-pages-ACTION-get-a-state-park-pass.html) 

We have the ability to have selectors for things like the following:

| Thing to target  | Selector |
| ------------- | ------------- | 
| A quick action link  | `.trk-quick-actions .trk-decorative-link`  | 
| A related action link  | `.trk-related-action .trk-decorative-link`  | 
| Any contact link  | `.trk-contact-group a`  |
| Contact phone number  | `.trk-contact-group .trk-contact-link-phone`  |

**note** I'll set up a time to do a proper code review all together, but feel free to leave thoughts here in the meantime if you'd like.

### DIFF
--- a/styleguide/source/_patterns/01-atoms/decorative-link.twig
+++ b/styleguide/source/_patterns/01-atoms/decorative-link.twig
@@ -1,7 +1,7 @@
 <span class="ma__decorative-link">
   {% if decorativeLink.type == "external" %}
-    <a href="{{decorativeLink.href}}" class="js-clickable-link">{{decorativeLink.text}}{% if decorativeLink.info %}<span class="ma__decorative-link__info">{{ decorativeLink.info }}</span>{% endif %}&nbsp;{% include "@atoms/05-icons/svg-external-link.twig" %}</a>
+    <a href="{{decorativeLink.href}}" class="js-clickable-link trk-decorative-link">{{decorativeLink.text}}{% if decorativeLink.info %}<span class="ma__decorative-link__info">{{ decorativeLink.info }}</span>{% endif %}&nbsp;{% include "@atoms/05-icons/svg-external-link.twig" %}</a>
   {% else %}
-    <a href="{{decorativeLink.href}}" class="js-clickable-link">{{decorativeLink.text}}{% if decorativeLink.info %}<span class="ma__decorative-link__info">{{ decorativeLink.info }}</span>{% endif %}&nbsp;{% include "@atoms/05-icons/svg-arrow.twig" %}</a>
+    <a href="{{decorativeLink.href}}" class="js-clickable-link trk-decorative-link">{{decorativeLink.text}}{% if decorativeLink.info %}<span class="ma__decorative-link__info">{{ decorativeLink.info }}</span>{% endif %}&nbsp;{% include "@atoms/05-icons/svg-arrow.twig" %}</a>
   {% endif %}
 </span>
--- a/styleguide/source/_patterns/02-molecules/contact-group.twig
+++ b/styleguide/source/_patterns/02-molecules/contact-group.twig
@@ -1,4 +1,4 @@
-<div class="ma__contact-group">
+<div class="ma__contact-group trk-contact-group">
   <h6 class="ma__contact-group__name">
     {% include contactGroup.icon %}
     <span>{{contactGroup.name}}</span>
@@ -11,13 +11,13 @@
 
       {# Online wrap value in a link #}
       {% if item.type == "online" %}
-        <a href="{{item.link}}" class="ma__content-link">{{item.value}}</a>
+        <a href="{{item.link}}" class="ma__content-link trk-contact-link-online">{{item.value}}</a>
       {# Phone - add tel: to href #}
       {% elseif item.type == "phone" %}
-        <a href="tel:{{item.link}}" class="ma__content-link ma__content-link--phone">{{item.value}}</a>
+        <a href="tel:{{item.link}}" class="ma__content-link ma__content-link--phone trk-contact-link-phone">{{item.value}}</a>
       {# Email - add mailto: to href #}
       {% elseif item.type == "email" %}
-        <a href="mailto:{{item.link}}" class="ma__content-link">{{item.value}}</a>
+        <a href="mailto:{{item.link}}" class="ma__content-link trk-contact-link-email">{{item.value}}</a>
       {# Address - RTE version of value and look for directions link #}
       {% elseif item.type == "address" %}
         <div class="ma__contact-group__address">
@@ -28,7 +28,7 @@
         {% endif %}
       {# Default - just spit out the value #}
       {% else %}
-        <span class="ma__contact-group__value">{{item.value}}</span>        
+        <span class="ma__contact-group__value">{{item.value}}</span>
       {% endif %}
     </div>
     {% if item.details %}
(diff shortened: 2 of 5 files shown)


jesconstantine (commented):
Can someone help me understand the difference, in the context of our tracking problem space, between: `.trk-quick-actions .trk-decorative-link` and `.ma__quick-actions a`?  Doesn't `.ma__quick-actions a` allow us to distinguish quick action from related action (which would be `.ma__related-action a`)?  Or am I misunderstanding the problem space and we really want to be distinguishing two other patterns?

Also - food for thought: 
I can see a future where we might not have a quick action or related action organism, but rather a generic organism like "linked list with heading" which can be implemented as a "related action" or "quick action" or "anything-else-the-content-team-might-dream-up" sidebar "widget".  And in that case, the difference is content/implementation and not semantic / available in pattern markup out of the box.  So then how would we know when to assign various tracking classes to each implementation of a generic pattern? 

I keep coming back to having some flag in the implementation of these patterns, which would mean passing a variable like Jonathan suggested yesterday.  For example, an "isTracked" boolean, and then maybe a "trackingName" string which is implemented as a data attribute or classname.

We could still provide some example implementations that end up being quite similar to what @isaacchansky  has already done, but it's passed to the pattern template in the json data file for the page, which is the implementation level of PL.

gender404:
I can answer the first question.  using the trk- prefix help separate concerns.

.ma__quick-actions => would be for CSS
.js-accordion => would be for JavaScript
.trk-sidebar-widget => would be tracking

This then allows us to mix and match class names without accidentally appling CSS, JS, or Tracking hooks.
<div class="ma__quick-actions js-accordion trk-sidebar-widget"></div>

If I don't want it to be tracked anymore, I can simply remove that class.  If we used .ma__quick-actions then I would have to update the controls in the Analytics tool.  Also if we decide to rename the component or switch the styling to ma__related-links, the tracking code doesn't have to change.  Does that make sense.

This is part of SMACSS if you want to read more about it

gender404 (approved):
This looks good to me

isaacchansky:
@jesconstantine I think what you're getting at is the over-arching concern of "since we're building re-usable components, if someone reuses two of the same thing, but wants them tracked differently, what do we do?" This is where our tracking scheme might be at odds with out development pattern/design language. Which I think is OK, but it's possible we'll need a second-level way to explicitly assign a unique tracking key to that component. The issue we started scratching our heads at, is that if that key is dynamic (i.e from drupal), we need a way to be positive that future deployments won't change it, and that it'll be consistent between environments. So that being set up by drupal will present these concerns. 

Hopefully 80% of these issues will be solved by being able to namespace a thing we want tracked by a parent organism, but in the case of two sidebar widgets as you mentioned, I think that's still an open question. 

The simplest solution I can come up with here (I think...) is that if we can't find a way to have unique tracking keys for multiple generic organisms that can't otherwise be distinguished *without* having to have authors involved or somehow tie it to content (authors still involved but not directly), then we have an issue of needing to update the tracking selectors whenever something changes on the page. So, with that being said, it's maybe easier to just use things like sibling selectors/first-child/etc in the tracking code to get what we need beyond the `.trk-` classes, since either way we do this google tag manager maybe has to be updated if people change the page. Having a simple test suite for google tag manager that is something like "for each selector, assert that it returns DOM node(s) on the page it's being applied to" would handle alerting us to where we need to do this, I think.


**concrete example**

we have a structure like this:
```
body
    main
    aside
        .trk-quick-actions
            .trk-decorated-link
        .trk-related-action
            .trk-decorated-link
        .trk-related-action
            .trk-decorated-link
        .trk-related-action
            .trk-decorated-link
```
So if we want to get at that second related action specifically, we can either assume that Drupal adds some sort of UUID and do: `#drupal-trk-id.trk-related-action .trk-decorated-link` OR have drupal not know or care about tracking at all, and do something like `.trk-related-action + .trk-related-action .trk-decorated-link`. If we have tests to ensure that each selector maps to actual DOM, we'd see something fail on "action page X" and then know to go adjust it.


**tldr;** I think relying on ordering in selectors as the gross thing might be better than relying on drupal UUIDs as the gross thing...


Other issues I have with a tracking flag via Drupal, is that in order for us to track a new conversion, we would have to ensure that: 1. the flag is set, then 2. caches are cleared respectively (if we eventually go to a static-site implementation as has been mentioned previously as a potential end-goal, we'll have to re-build) and then 3. ensure that we have no regressions in the future since this is in content, not code... If we already have everything able to be tracked by default, then it's just a google tag manager change & we're live. Let me know if I'm thinking about this correctly.

gender404:
Could the problem be solved using JS/jQuery leveraging the index of the element found and appending that before sending it to GTM (.trk-quick-action-1)?

This, however, made me wonder if we should be using a more generic class to hook into.  Like adding "trk-parent" and "trk-link" to items we want to track and a corresponding data attribute with more information.  This would make the JS needed much simpler.

~~~
<div class="trk-parent" data-trk-info="quick-actions">
  <a class="trk-link" data-trk-info="decorative-link">learn more</a>
</div>
~~~

~~~
var GTM = [];

$('.trk-parent').each(function(index){
  var parent-id = $(this).data('trk-info') + index;

  $(this).find('.trk-link').each(index) {
    GTM.push( parent-id + " - " + $(this).data('trk-info') + " - " + index );
  });
});
~~~

gender404 (changes requested):
see latest comment

GBirch:
I may be missing the point, but I think I disagree with Isaac.  Seems to me that whatever engine is using these patterns (Drupal, WordPress, Jekyll, you name it), you are going to want to be able to enable tracking of specific items of content.  In short, at least as I understand it, it really IS content we are trying to track.   Relying on the order in which items appear on the page seems inherently fragile (updating GTM whenever an author decides to re-order the links on a page seems pretty gross to me, and likely to be a constant problem).  
On the Drupal side, UUIDs are the obvious, built-in mechanism for identifying content, but by no means perfect.  With UUIDs, we're going to have problems whenever an author decides to change content by deleting the old node and creating a new one, which is NOT a merely theoretical problem.  And UUIDs will often not be consistent across environments (although I don't understand why that matters - what environments other than live are you trying to track?).  But that's not a Mayflower problem, it's a problem to be solved by the content management system, whatever it is.  As far as Mayflower is concerned, why wouldn't we output and use an identifier (and I think @legostud 's suggesting for a tracking info attribute is nicely flexible), wherever it comes from and however it is generated?

isaacchansky:
@GBirch all good points. Are Drupal UUIDs at the paragraph/atomic level? & how stable would they be? If someone re-arranges paragraphs on a page, are the UUIDs constant? 

To clarify, the uniqueness requirement for these things is within a single page. There's no benefit to have thoroughly unique ids across the site. 

I'm also not sure how real of a problem needing to distinguish between two of the 'same' items that are adjacent to eachother (i.e the order-dependent selectors might be a fix for a non-problem). This might require another conversation with the data team to firm up the concrete requirements.

gender404:
Any objections if I close this Pull Request?

isaacchansky:
I say go ahead an close it. If this works get prioritized again someday it won't be hard to reopen the conversation.

PR #261, opened 2016-12-22
smooth header gradient

isaacchansky:
**jira**: https://jira.state.ma.us/browse/MQA-198

The core issue here is that the banner gradient is not perfectly @ 180 degrees, it's 150 or something. So, having a solid block to fill in the left gap doesn't work seamlessly. 
Not in love with my solution since I'm adding a one-off breakpoint, but it seeeeeeems to me to be working across all breakpoints.

### DIFF
--- a/styleguide/source/assets/scss/03-organisms/_page-banner.scss
+++ b/styleguide/source/assets/scss/03-organisms/_page-banner.scss
@@ -22,15 +22,20 @@ $page-banner-small: '.ma__page-banner--small';
     font-size: 1rem;
     max-width: 1050px;
     padding: 30px 25px 45px 40px;
-    left: -20px; // account for container margin to be flush to screen-left
+    left: 0px; // account for container margin to be flush to screen-left
     position: absolute;
     top: 50%;
     transform: translateY(-50%);
     z-index: 1;
 
+    @media (min-width: 1320px) {
+      // Account for all space to the left of the 1320-max-width container, plus the gutter.
+      padding-left: calc(((100vw - 1320px)/2) + 20px) !important;
+    }
+
     @media ($bp-x-small-min) {
       display: flex;
-        align-items: center;
+      align-items: center;
       margin-right: 20px;
       padding-top: 50px;
       padding-right: 175px;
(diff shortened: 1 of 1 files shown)


styleguide/source/assets/scss/03-organisms/_page-banner.scss
```diff
@@ -21,38 +21,35 @@ $page-banner-small: '.ma__page-banner--small';
   &__content {
     font-size: 1rem;
     max-width: 1050px;
-    padding: 30px 25px 45px 0;
+    padding: 30px 25px 45px 40px;
+    left: 0px; // account for container margin to be flush to screen-left
     position: absolute;
```
gender404, line 26: I prefer to group "left" under position: absolute with an indent.  It helps with readability.  Also you don't need units if the value is 0.

position: absolute;
  top: 50%;
  left: 0;

styleguide/source/assets/scss/03-organisms/_page-banner.scss
```diff
@@ -21,38 +21,35 @@ $page-banner-small: '.ma__page-banner--small';
...
+    @media (min-width: 1320px) {
+      // Account for all space to the left of the 1320-max-width container, plus the gutter.
+      padding-left: calc(((100vw - 1320px)/2) + 20px) !important;
+    }
+
     @media ($bp-x-small-min) {
       display: flex;
-        align-items: center;
+      align-items: center;
       margin-right: 20px;
```
gender404, line 39: the indentation was intentional.  Align-items is related to display:flex so it's my way of grouping dependent items without breaking the arrange alphabetical rule for attributes.

gender404 (changes requested):
There are some issues with this solution.  It does solve the request, but introduces some other issues.

1. The left padding is not 40px.  It changes from 40 -> 30 -> 20

2. When the browser is extra wide the green gradient could be transparent beyond where we wanted it to end.

3. Using VW units is tricky, because it includes the scrollbar.  I'm not sure you will be able to line up the content of this banner with the content on the page itself.

I don't have a better solution yet, but I can see one floating around in my head. 

I like the solution for below the max-width assuming we fix the padding issue, which could be solved by Including ma-container and overriding the right padding it adds.  For max-width, maybe we leave the :before element, but adjust when the gradient switches to full green linear-gradient(... , $c-theme-green 90%).  Not sure how well that will work.

gender404:
I thought about this overnight a I think I have a simple solution.
1. convert the gradient to 180deg so it ends at the solid color.
2. skew the __content object 30deg to get the desired angle (I think this will skew the :before as well)
3. skew the title the opposite way to put it back to normal

isaacchansky:
This was solved in PR #264 , closing.

PR #263, opened 2016-12-22
312 section divider

gender404:
New Section divider atom


gleroux02 (approved):
Everything looks good here.

PR #264, opened 2016-12-23
MQA-198 - Page Banner - fixing blending issue with background

gender404:
This should resolve your issue with MQA-198.

If it does, please feel free to merge it into the repo


isaacchansky:
I really like this solution, nice 👍

isaacchansky (commented):
This feels a lot cleaner & as you mentioned also solves the fuzzy-edge issue. Awesome.

PR #265, opened 2016-12-27
Fix top-level link JS in main navigation

isaacchansky:
Remove `preventDefault` that was stopping the link-following behavior of top-level links.

**Jira:** https://jira.state.ma.us/browse/MQA-195

### DIFF
--- a/styleguide/source/assets/js/modules/mainNav.js
+++ b/styleguide/source/assets/js/modules/mainNav.js
@@ -158,7 +158,6 @@ export default function (window,document,$,undefined) {
     });
     $mainNavToggle.children('a').on('click', function(e) {
       if(windowWidth <= breakpoint) {
-        e.preventDefault();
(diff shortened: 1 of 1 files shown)


styleguide/source/assets/js/modules/mainNav.js
```diff
@@ -158,7 +158,6 @@ export default function (window,document,$,undefined) {
     });
     $mainNavToggle.children('a').on('click', function(e) {
       if(windowWidth <= breakpoint) {
-        e.preventDefault();
         let $content = $(this).parent().find('.js-main-nav-content');
```
gender404, line 161: I'm pretty sure this is required in order to open the submenu on Mobile.  Only top level links with menus should be considered "$mainNavToggle".

isaacchansky: yeahh you're right, got confused since links are just '#' in PL. gonna scrap this.

PR #266, opened 2016-12-27
Fix markup for contact group

isaacchansky:
- adjust headlines to be consecutive
- align css with proper sizing decoupled from markup

**Jira:** https://jira.state.ma.us/browse/MQA-187

### DIFF
--- a/styleguide/source/assets/scss/01-atoms/_column-heading.scss
+++ b/styleguide/source/assets/scss/01-atoms/_column-heading.scss
@@ -1,6 +1,3 @@
 .ma__column-heading {
   @include ma-h5;
-  font-size: 1.625rem;
-  line-height: 1.3;
-  margin-bottom: .5em;
 }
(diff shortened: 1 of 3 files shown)


styleguide/source/assets/scss/01-atoms/_column-heading.scss
```diff
@@ -1,5 +1,6 @@
 .ma__column-heading {
+  @include ma-h5;
   font-size: 1.625rem;
   line-height: 1.3;
   margin-bottom: .5em;
-}
```
gender404, line 5: The @include ma-h5 is adding all the same styles that follow it.  Please remove the font-size, line-height, and margin-bottom from this selector.  It should just be:

.ma__column-heading { 
@include ma-h5;
}

gender404 (changes requested):

styleguide/source/assets/scss/01-atoms/_column-heading.scss
isaacchansky: hah, fuzzy brain today, thanks!

PR #272, opened 2016-12-29
Fixes MQA-235-video-title

powpow12:
Fixes 
- https://jira.state.ma.us/browse/MQA-235

Note:  Not 100% how the js object plays with the twig file.  I see it as a placeholder for what we expect drupal to pass to the template so I wanted to makes sure that I included a new property `video.title`.  That being said, even after a new `gulp build`, I couldn't get the title to populate.  Please let me know if I got this right.


gender404:
This change should be populating the title at the Atom level.  I'm guessing you're not seeing it change at the Page Level, because the JSON file has to be updated on the page to add the title there as well.

powpow12:
touche, thanks.

PR #275, opened 2016-12-30
Add explicit aria label to link nav with headline

isaacchansky:


styleguide/source/_patterns/03-organisms/by-template/jump-links.twig
```diff
@@ -1,8 +1,8 @@
 <section class="ma__jump-links">
   <div class="ma__jump-links__inner">
     <div class="ma__jump-links__container">
-      <h2 class="ma__jump-links__title">{{jumpLinks.title}}</h2>
-      <nav class="ma__jump-links__links js-scroll-anchors">
+      <h2 class="ma__jump-links__title" id="guide_contents">{{jumpLinks.title}}</h2>
+      <nav class="ma__jump-links__links js-scroll-anchors" aria-labelledby="guide_contents">
         <button class="ma__jump-links__toggle-link js-scroll-anchors-toggle">+</button>
```
gender404: Let's not use "guide_contents" since these jump links are used for stacked row content which will be a lot more pages than just "guides".  Maybe we just call it "page-contents" or something similar to what you use for the jump links in the Action detail component.

isaacchansky: Makes sense, thanks

PR #276, opened 2016-12-30
MQA-64 contact accordion implementation refactor

jesconstantine:
WIP PR to refactor contact us accordion implementation, making it so that the first two groups of contact information for an entity always appear and any subsequent group is hidden in a drop down accordion.

### Ticket
Address mayflower code aspects of: https://jira.state.ma.us/browse/MQA-64

### Notes
Prior implementation has the contact us molecule pattern looking for a "hidden" attribute passed as a variable to the template and associated with each contact group.  This would imply that there needs to be input from a content author as an implementation detail (see conversation here: https://github.com/massgov/mayflower/pull/189/files#diff-80ce4390df2666783f866d9b5c81ef06R15 )

This PR eliminates the "hidden" attribute, since the need for the accordion can be determined and applied automatically based on the length of the contact groups array.

### Implementation / Integration
There will still need to be some implementation changes in massgov/mass.  I'm happy to continue that work, but would like some insight from @nstriedinger as I believe he has done existing sidebar / contact work (which is why I've asked him to review this PR)

### DIFF
--- a/styleguide/source/_patterns/02-molecules/contact-us.twig
+++ b/styleguide/source/_patterns/02-molecules/contact-us.twig
@@ -1,4 +1,3 @@
-{% set accordion = false %}
 <section class="ma__contact-us js-accordion">
   {% if contactUs.title %}
     <h4 class="ma__column-heading">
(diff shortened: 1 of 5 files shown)


styleguide/source/_patterns/02-molecules/contact-us.twig
```diff
@@ -11,9 +11,12 @@
...
   {% endif %}
 
+  {# If more than 2 contact groups, they should load inside of accordion.#}
   {% for contactGroup in contactUs.groups %}
-    {% if contactGroup.hidden and not accordion %}
+    {% if loop.length > 2 %}
       {% set accordion = true %}
+    {% endif %}
+    {% if loop.index == 3 %}
       <div class="ma__contact-us__extra js-accordion-content">
```
gender404: This isn't very optimal.  You have two if statements within a for loop.  Those if statements will be check on each instance of the loop when they don't really need to.  I'm not even sure we need them at all.

I would suggest doing two for loops instead.  
1. Modify the existing loop: contactUs.groups | [start:2] so it only does the first two
2. After that loop, insert a conditional based on contactUs.groups | length > 2
3. inside that conditional is the ma__contact__us__extra element
4. inside that element is another loop contactUs.groups | [2:last] to render the remain contact groups.

Similar Example:
https://github.com/massgov/mayflower/blob/dev/styleguide/source/_patterns/03-organisms/by-template/jump-links.twig

jesconstantine: Thanks @legostud -- this makes a lot of sense, and sorry I missed it the first time.  Refactored in https://github.com/massgov/mayflower/pull/276/commits/a9d4fdbc9954ac338e2e9e822e9a07a6f33097a1

PR #277, opened 2017-01-02
Remove group class and apply style to ma__contact-group__name

nstriedinger:


gender404:
I have some issues with this change.  Is there a reason why we are trying to remove the outer div?

The biggest red flag I see here is that you've broken the "contact-group" BEM syntax which makes wonder if this component needed to be refactored into smaller molecules.

BEM requires a parent (ma__contact-group) wrapping all of the related child (ma__contact-group__name)

gender404:
I looked at the PR #265 and what didn't match in the screen shot doesn't match the changes made here so I'm still confused by the changes made.

gleroux02:
I'd like to think about this a bit further too. I'm not comfortable gutting this as much as was needed because of the implementation problems in drupal.

gleroux02:
I opened up a PR https://github.com/massgov/mass/pull/309 on the drupal side that adds some css override and keeps the markup as is. I don't think it is something that should be ported to the styleguide as it is an override for the way we implemented the paragraph setup.

PR #282, opened 2017-01-04
MGRP-344 - rich text image/figure

gender404:
1. Created new atom "figure".
  * includes the image atom
  * uses 'align-left' and 'align-right' classes to display at 50% width and floated left or right.
2. Modified the Action Image molecule to use this new atom
3. Modified the Image atom to remove the pull-left pull-right classes.
4. Updated other patterns to use the new Figure atom instead of the Image atom.


styleguide/source/assets/scss/01-atoms/_figure.scss
```diff
@@ -0,0 +1,38 @@
+.ma__figure {
+  margin: 0 0 1.75rem;
+
+  @media ($bp-small-min) {
+    &.align-left {
```
gleroux02, line 5: Can you additionally add the inline-left and inline-right classes or switch from using align-left? This will be a help when the cms is drupal as the classes added to inline images in the WYSIWYG are incredibly difficult to alter.

gleroux02 (dismissed):

PR #285, opened 2017-01-05
Dp 708 org page

gender404:
1. New Org Page created to demo the page type in the comps
2. Icon Links - removed the Sidebar Heading
3. SectionThreeUp 
  - Fixed bug with title not rendering on L1 page
  - Fixed padding issues
4. Page Banner
  - Centered background image
  - Added new large variant (breaking change => 'small' in now 'size')
  - Added new titleSubText
5. Action Header
  - Modified to include optional Page Banner 
  - Modified to include optional divider
  - refactored how widgets were being included


release-notes.md
```diff
@@ -1,5 +1,11 @@
 # Mass Gov Release Notes
 
+## 3.1.0 (Sprint 10)
```
gleroux02, line 3: I'd be interested in this change log including references to changes to existing data structures. In this case there is a change to the page-banner data structure: https://jira.state.ma.us/browse/DP-926.

styleguide/source/_patterns/03-organisms/by-template/action-header.twig
```diff
@@ -1,20 +1,29 @@
...
-  <img aria-hidden="true" alt="icon-sprite" src="/{{ directory }}/images/svg-loc-icons.svg" style="display:none" />
-  <div class="ma__action-header__inner">
+{% if pageBanner %}
+  {% set bannerClass = "ma__action-header--banner" %}
+{% endif %}
+
+<section class="ma__action-header {{ bannerClass }}">
+  {% if pageBanner %}
+    <div class="ma__action-header__banner">
+      {% include "@organisms/by-template/page-banner.twig" %}
```
gleroux02: This is the only place where we are rendering the page-banner template within a different organism. On other pages it is a stand alone element added to the json of the page and rendered by calling the page-banner organism on the page template. Would we be able to remove the rendering of this from action-header and have it in the org page twig file instead? It would require altering how the spacing between the banner and the ma__action-header__content is figured out

gleroux02 (commented):
@legostud I have two open questions for you. I'm not 100% on the page-banner question but I want to try to keep the rendering of elements consistently throughout the styleguide when applicable.

release-notes.md
gender404: I'm going to push up a change to correct the breaking change issue with Page Banner.  I think I can make it backward compatible.  It just won't be as clean as I like it.

gender404: x.1.x is a minor release (new features).  I can add notes though when there is a 1.x.x major release though as to why it was changed going forward.

styleguide/source/_patterns/03-organisms/by-template/action-header.twig
gender404: I thought about this this morning as well.  It makes more sense to un-nest the page-banner from the action-header.  I can make that change.  I just wasn't sure if that was Kosher in Drupal to have two page headers.

release-notes.md
gleroux02: I think it is fine to have a breaking change. As long as we have a ticket that covers the change and it's documented. I'd rather have the clean approach when possible.

gender404: I remember why I didn't make it a breaking change (4.x), I already added a ticket to this sprint to fix the issue in Drupal.  https://jira.state.ma.us/browse/DP-909

Hopefully that can get resolved easily this sprint.

styleguide/source/_patterns/03-organisms/by-template/action-header.twig
gender404: I just pushed up the change for this.

release-notes.md
gleroux02: Ah didn't see that ticket. I'll close mine as a duplicate.

styleguide/source/_patterns/03-organisms/by-template/action-header.twig
```diff
@@ -1,13 +1,4 @@
...
-<section class="ma__action-header {{ bannerClass }}">
-  {% if pageBanner %}
-    <div class="ma__action-header__banner">
-      {% include "@organisms/by-template/page-banner.twig" %}
-    </div>
-  {% endif %}
+<section class="ma__action-header">
   <div class="ma__action-header__container">
     <div class="ma__action-header__content">
       {% include "@organisms/by-template/page-header.twig" %}
```
gleroux02, line 4: I'm not liking the need to hide the page title with display: none. But I'm not sure what the solution should be. Any thoughts on this?

gleroux02: After having a conversation with @legostud we realized that there isn't a great way around this and it will have to do for now.

PR #290, opened 2017-01-09
DP-712 - Illustrated links - making background image optional

gender404:


styleguide/source/_patterns/02-molecules/illustrated-link.twig
```diff
@@ -6,5 +6,10 @@
       {% include "@atoms/decorative-link.twig" %}
     </div>
   </div>
-  <div class="ma__illustrated-link__image" style="background-image: url('{{illustratedLink.image}}')"></div>
+  <div 
+    class="ma__illustrated-link__image" 
+    {% if illustratedLink.image %}
```
gleroux02, line 9: I would prefer this to be a ternary operator so that it is easier to read, but I haven't been able to write one that also renders the image url properly. Was this something you tried and also had an issue with?

gleroux02 (approved):

gleroux02:
@legostud I'd love to make that if statement cleaner, but if you tried by using a ternary statement and also ran into the same problem as me then I'd rather just merge it on in.

styleguide/source/_patterns/02-molecules/illustrated-link.twig
gender404: I've had bad luck try to manipulate html tags in the past using a ternary operator.  It spit out the results as a string.  

Only other option I can think of would be to duplicate the ```<div>``` and use if/else statement to include the correct version (with or without the style attribute).

PR #292, opened 2017-01-10
Gulp - updating gulp-sass to work with Node version 6.x

gender404:
Our dev server (dev.mayflower.velir.com) was upgraded to Node 6.x.  This cause the gulp build to fail when trying to run gulp-sass.  I've upgraded gulp-sass to version 3.x and this appears to resolve the issue.

Note: I have not tested this on an older version of Node.  Running gulp will continue to use the older version of gulp-sass unless someone runs npm install to get the newer version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@ This living styleguide is built using PatternLab.
 
 ##Machine set up
 1. Install PHP
-2. Install NodeJS (https://nodejs.org/en/download/)
+2. Install NodeJS version 6.9.4 (https://nodejs.org/en/download/)
 3. Install GulpJS, via command line `npm install -g gulp`
 
 ##Set up instructions
--- a/styleguide/npm-shrinkwrap.json
+++ b/styleguide/npm-shrinkwrap.json
@@ -1,0 +1,6349 @@
+{
+  "name": "MayflowerPatternLibrary",
+  "version": "2.0.0",
+  "dependencies": {
+    "abbrev": {
+      "version": "1.0.9",
+      "from": "abbrev@>=1.0.0 <2.0.0",
+      "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.0.9.tgz",
+      "dev": true
+    },
+    "accepts": {
+      "version": "1.3.3",
+      "from": "accepts@>=1.3.3 <1.4.0",
+      "resolved": "https://registry.npmjs.org/accepts/-/accepts-1.3.3.tgz",
+      "dev": true
+    },
+    "acorn": {
+      "version": "3.3.0",
+      "from": "acorn@>=3.1.0 <4.0.0",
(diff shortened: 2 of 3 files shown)


jesconstantine (changes requested):
@legostud - thanks for recognizing this as an abnormal change and flagging so many reviewers.

I was able to build successfully after `npm-install` to update gulp-sass, here is my node version:
image

However, since I don't have enough of an understanding about what versions / environments folks who build mayflower locally are using (and this will remain the case as this population will only grow in the future), I'd ask as that as a part of this PR (and all PR's of this nature), we update both the release notes the readme for this repository with any new environment constraints / requirements.

I didn't see anything directly in release notes for gulp sass: https://github.com/dlmanning/gulp-sass/releases but following links to release notes for node-sass, I see a table of supported environments which looks useful: https://github.com/sass/node-sass/releases/tag/v4.2.0

gleroux02:
@legostud I think we should also be using npm shrinkwrap to help pin down the versions of npm and gulp packages https://docs.npmjs.com/cli/shrinkwrap. I know it is something that the engineers here were really pushing for our internal styleguide tool. It may help here too.

gleroux02 (approved):
I was able to update without issue and run gulp. We can't say that this will 100% work across everyone's environments but I feel pretty good about it. I'll suggest that we should add npm-shrinkwrap to help lock down versions for everyone.

gender404:
doesn't the package.json lock down the packages?

gleroux02:
@legostud shrink wrap worked just fine. @jesconstantine would you be able to pull changes and run npm install and make sure you can still run the styleguide?

jesconstantine:
@gleroux02 @legostud - I was also able to build with the addition of npm shrinkwrap package.  I had a coworker test (he's on node v7.*) and he could build as well.

I have two questions:
1. Will read the docs more deeply, but at glance it seems that shrinkwrap.json serves a similar purpose as a composer.lock file -- does that sound right?
2. I get that adding shrinkwrap will help us lock down npm package versions (which I think is necessary and great), does it also help us with the question posed in this PR, which I believe to be how our packages interact with node versions?

cc: @isaacchansky thought you'd find this thread interesting, might have some thoughts as well.

gleroux02:
@jesconstantine yes shrinkwrap will act just like composer.lock. It will lock down all modules and sub modules to a specific version at a given period of time. This won't solve knowing how specific versions interact with a given version of node itself. However we probably should include in the readme what version of node the library is intended to work with. If we have that documented and then shrinkwrap set then it would be on the user to make sure they are working with the version of node that was set in the readme.

jesconstantine:
Thanks @gleroux02 -- I agree: documenting that node version dependency for developers is the only thing that I think this PR is missing.

jesconstantine (approved):

jesconstantine:
Thanks @legostud I was just about to do that.

PR #293, opened 2017-01-11
Set Mayflower default pattern to greeting / readme page

jesconstantine:
This PR creates a (sparse) homepage or Mayflower, which currently displays the version, date, and a greeting.

This page should be more of a readme than a greeting, but this is a start. (Please feel free to add more information, suggest better patterns, plan for the future, etc.)

Would love to add to the deploy process a step where the version gets bumped and the contents of this page are updated.

### Screenshot of current page content
image

### DIFF
--- a/styleguide/source/_patterns/05-pages/Readme.json
+++ b/styleguide/source/_patterns/05-pages/Readme.json
@@ -1,4 +1,9 @@
 {
+  "narrowTemplate": {
+    "side": "left",
+    "color": "blue"
+  },
+
   "errorPage": {
     "type": "v2.2.1 (01/09/2016)",
     "title": "Welcome to Mayflower, the Commonwealth's Design System",
--- a/styleguide/source/_patterns/05-pages/Readme.json
+++ b/styleguide/source/_patterns/05-pages/Readme.json
@@ -1,7 +1,7 @@
 {
   "narrowTemplate": {
     "side": "left",
-    "color": "blue"
+    "color": "green"
   },
 
   "errorPage": {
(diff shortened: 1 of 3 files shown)


gender404:
Looks good to me.  I added the green diagonal background to this.  

image

styleguide/source/_patterns/05-pages/Readme.json
```diff
@@ -0,0 +1,12 @@
+{
+  "narrowTemplate": {
+    "side": "left",
+    "color": "green"
+  },
+
+  "errorPage": {
+    "type": "v2.2.1 (01/09/2016)",
```
isaacchansky, line 8: Is this the only place we update the Version number? I'm thinking about keeping it in sync if it's in more than one spot...

isaacchansky (commented):
Looks good! nice work. I'll defer to Harlan, but I could see us maybe wanting to drop the background color if this gets more text-heavy, for readability.

gender404 (approved):
If we get more content to put in here, we might want to get the design team involved.  For now, I think this is a good start.

styleguide/source/_patterns/05-pages/Readme.json
jesconstantine: It's the only place to my knowledge.  @legostud - can you think of any other place?

gender404: the release-notes.md file in the repo is the only other place I can think of.

jesconstantine: Oh, yes, right.  That makes sense. :)

At any rate, @isaacchansky this speaks to my desire to automate.  But I thought this was rather MVP...

isaacchansky: yep! totally agree, I like it as is.

gender404:
this ticket seems quite now so I'm assuming it's fine to merge this request even though the reviewers haven't all submitted an approval.

jesconstantine:
I agree: So I'm game if either @legostud or @isaacchansky want to merge.

PR #298, opened 2017-01-13
DP-787 park icons

gender404:
1. Added new SVG icon files
2. Removed Mountain icon file
3. Modified Gulp to stop building the location page SVG sprite file.
4. Updated location icon twig files to contain the SVG code directly (better matches actual usage)
5. Create new "svg2twig" gulp task to auto generate twig versions of the svg icon files
6. moved all svg icon files into a new "svg-icons" folder
7. created a symbolic link between svg-sprite to svg-icons folders
* attempting to make this change backwards compatible.
* not 100% sure this will work with the Drupal implementation

Other fixes
1. Removed duplicate clock.svg icon by using wait time icon for callout time organism
2. updated release notes to reflect the scope of sprint 10
3. Set pattern lab to not show documentation by default on 'view all' pages
4. Added some real links to the main navigation organism

https://jira.state.ma.us/browse/DP-787


lukewertz:
@gleroux02 I'd like to see us stop assuming where icons are going to be used. In a modular system, I expect components (even icons) to be re-usable in different locations. I don't think this artificial construct of a "location page" warrants its own icons.

cc @legostud

gender404:
@lukewertz - I'm all for re-usable.  In this case we have a component "location icons" that is used to tell the visitor whether or not a location includes certain features.  For MVP, there are up to four features and eventually there will be up to eight of these displayed.  As such, I've put them in a separate "locations" folder to better document that these icons are the ones used to describe a location.

The "location icons" component probably should be renamed, but that would be part of a larger refactor effort to avoid making breaking changes.  We also need to refactor most of the svg icons in Pattern Lab to determine if they need to be part of an SVG sprite or to include the SVG code directly.

allen-black:
Hey all. We need these icons to be available to content editors and site builders to use in taxonomy, which is what surfaces them to end users. For these icons to be usable, we need to adjust the mechanism for icons being available in Drupal so different folders can be used or we need to move these to /massgov/mayflower/tree/dev/styleguide/source/assets/images/svg-sprite , where the existing mechanism is. I have no opinion, but looking for a way to state if the location is wheelchair accessible

gender404:
I've refactored how icons work in Pattern Lab.  Originally we were using an svg-sprite to bundle all the svg files in the svg-sprite folder.  Now we are just converting the svg files into twig files that can be included directly in Patterns, which matches the implementation in Drupal.

gleroux02 (changes requested):
I'd like to merge 06-icons-location, 07-user-added-icons into the 05-icons directory. That way we would only have one directory and would simplify implementation. Also we can remove the "Deprecated" state for the icons as this seems like it would be a solution we can continue with.

gender404:
1. I removed the Deprecated state and any related copy from the markdown files.
2. I removed the locations folder and put all the files into the main icon folder.  I had intended to do that originally, but I guess I got distracted.

The is another ticket for this sprint around funnel Icons, which are the user added icons.  I will remove the user added folder as part of that ticket.

gleroux02:
Would it make sense to add a sym link of 06-icons-location to 05-icons to allow for backwards compatibility? And the same for the user added directory when that is removed in the other ticket?

gender404:
not sure if that would work in Pattern Lab.

gleroux02:
I'll verify that we don't need to worry about this. And now that I think about it further it may not be an appropriate solution anyways.

gleroux02 (approved):

PR #302, opened 2017-01-18
DP-1103 - Downloads - generic icon

gender404:


gleroux02:
@legostud should we wait until #298 is in and then refactor this to work with it?

gender404:
icon files have been updated to use the new pattern.

PR #303, opened 2017-01-18
DP-1278 - RTE external links - correcting alignment

gender404:
Alignment of links in the Rich Text was incorrect after JS converted them to decorative links (external links).

https://jira.state.ma.us/browse/DP-1278


gleroux02:
@legostud where is the best place within the styleguide to check on this?

gender404:
try here: http://dev.massgov.velir.com/?p=pages-GUIDE-movng-to-ma-part1

PR #304, opened 2017-01-19
DP-1267 structure data page elements

ygannett:
Added Schema.org markup for structured data to improve external search.

The changes in this branch is common to all pages in the site.

* Organization infomration as site owner's information
   - JSON-LD
* Page elements identifiers for search engines
   - JSON-LD for a list of the elements
   - RDFa for identifing each area of the list in <body>
* A color state seal image file to be used for Information Graph panel in Google search results page as the organization logo


styleguide/source/_meta/_00-head.twig
```diff
@@ -15,9 +15,65 @@
...
+  ORGANIZATION
+  - alternateName can be added later if helpful.
+#}
+    <script type="application/ld+json">
+      { 
+          "@context" : "http://schema.org",
+          "@type" : "Organization",
+          "url" : "https://www.mass.gov/",
+          "logo": "https://www.mass.gov/assets/images/stateseal-color.png",
+          "name" : "Commonwealth of Massachusetts",
```
gender404: I'm thinking the domain here should be pulled from the current url, unless there is a need for it to say www.mass.gov.  This will make it easier to test before launching to prod.

```
"url": document.location.href;
"logo": document.location.href + "assets/images/stateseal-color.png"
```

styleguide/source/_meta/_00-head.twig
```diff
@@ -15,9 +15,65 @@
...
+  - footer area
+  - sidebar area
+  - main navigation area
+#}
+    <script type="application/ld+json">
+      {
+          "@context":"http://schema.org",
+          "@type":"WPHeader",
+          "@id" : "https://www.mass.gov//#header"
+      }
```
gender404: why // before the #header?

styleguide/source/_patterns/04-templates/two-column.twig
```diff
@@ -12,7 +12,7 @@
         {% include "@base/placeholder.twig" with {'placeholder':{'text':'Page content'}} %}
       {% endblock %}
     </div>
-    <aside class="sidebar">
+    <aside class="sidebar" id="sidebar">
```
gender404: This works here.  Will you need to add this to stacked-row pages that have multiple sidebars on the page?
http://dev.massgov.velir.com/?p=templates-stacked-row-template.  

IDs cannot be reused on the same page so "sidebar" wouldn't work as an ID there.

gender404 (changes requested):

styleguide/source/_meta/_00-head.twig
ygannett: I'll make those changes.  thank you.

styleguide/source/_meta/_00-head.twig
ygannett: I meant to put a note about this.  I'm sorry that I missed that.  It's supposed to have the page location.  I thought that needs to be done in Drupal, not in mayflower.  
From your comment for the domain one above, `"@id" : "https://www.mass.gov//#header"` can be replaced with `"@id" : "document.location.href/#header"`.  Does `document.location.href` point to only the domain part?

styleguide/source/_patterns/04-templates/two-column.twig
ygannett: It doesn't get validated with class in the testing tool (https://search.google.com/structured-data/testing-tool). I thought about using the existing classes, and even tried  `@class" : "https://www.mass.gov//.sidebar` with `aside class="sidebar" class="sidebar">`  Google doesn't recognize classes for an object.
To accommodate multiple <aside>s, we might need to:
1. limit the use to certain page types, 
2. use RDF-a only instead of JSON-LD, or 
3. omit this particular element.
I need to confirm the use case of your sample.  Thank you.

styleguide/source/_meta/_00-head.twig
gender404: Sorry, I made a typo.  This should be document.location.origin

styleguide/source/_meta/_00-head.twig
gender404: document.location.href contains the entire url, which sounds like what you want.

styleguide/source/_meta/_00-head.twig
ygannett: Ok.  Thank you for letting me know that.

styleguide/source/_meta/_00-head.twig
ygannett: Perfect!  Thank you.

styleguide/source/_patterns/04-templates/two-column.twig
ygannett: It's not ready for review yet.
Top two items have been changed, but I'm still working on this one.

ygannett:
I finished all my work.  
The markup for the sidebar has been removed to accommodate multiple `<aside>`s in a page.

ygannett:
I have a little concern about the concatenation.
For example, 
`` "logo": '"' + document.location.origin + '/assets/images/stateseal-color.png' + '"',
``
gets output as `"logo": "http://www.mass.gov/assets/images/stateseal-color.png",`?
When I check it locally, I don't see that way.  Just wanted to make sure if it needs more work...

styleguide/source/_meta/_00-head.twig
```diff
@@ -15,9 +15,83 @@
...
+  - add to home (front) page only.
+  - about can be added when its content is ready.
+  - keywords can be added when its content is ready.
+#}
+{% if schemaPageType == "Home" %}
+    <script type="application/ld+json">
+        {
+          "@context": "http://schema.org",
+          "@type": "WebSite",
+          "url": '"' + document.location.origin + '"',
```
gender404: document.location.origin returns a string so you don't need the '"' + ...  +'"'.  Adding that outputs ""https://massgov.com""

Should be => "url": document.location.origin,

styleguide/source/_meta/_00-head.twig
```diff
@@ -15,9 +15,83 @@
...
+          "name": "Mass.gov",
+          "alternateName": "The Official Website of the Commonwealth of Massachusetts",{#
+          "about": "~NEED CONTENT~", #}
+          "copyrightHolder": "Commonwealth of Massachusetts",{#
+          "keywords": "state government, ~OTHERS TO ADD: Keywords or tags used to describe this content. Multiple entries in a keywords list are typically delimited by commas.~", #}
+          "potentialAction": {
+             "@type": "SearchAction",
+             "target": '"' + document.location.origin + '/?s={search_term_string}&post_type=post' + '"',
```
gender404: document.location.origin returns a string so you don't need the '"' + ...  +'"'.  Adding that outputs ""...""

Should be => "target": document.location.origin + "/?s={search_term_string}&post_type=post"

I am curious about what this is => {search_term_string}.  If it's supposed to be twig variable, it would be {{ search_term_string }}, otherwise the value is just going to be the string "{search_term_string}"

styleguide/source/_meta/_00-head.twig
```diff
@@ -15,9 +15,83 @@
...
+  SITE LEVEL COMMON DATA: 
+  ORGANIZATION
+  - alternateName can be added later if helpful.
+#}
+    <script type="application/ld+json">
+      { 
+          "@context": "http://schema.org",
+          "@type": "Organization",
+          "url": '"' + document.location.origin + '"',
+          "logo": '"' + document.location.origin + '/assets/images/stateseal-color.png' + '"',
```
gender404: should be:
"url": document.location.origin,
"logo": document.location.origin + '/{{ directory }}/images/stateseal-color.png',

{{ directory }} is needed, because images are in a different location within drupal

styleguide/source/_meta/_00-head.twig
```diff
@@ -15,9 +15,83 @@
...
+          "@context": "http://schema.org",
+          "@type": "WPFooter",
+          "@id": '"' + document.location.href + '#footer' + '"'
+      }
+    </script>  
+    <script type="application/ld+json">
+        {
+          "@context": "http://schema.org",
+          "@type": "SiteNavigationElement",
+          "@id": '"' + document.location.href + '#main-navigation' + '"'
```
gender404: should be => "@id": document.location.href + '#main-navigation'

styleguide/source/_meta/_00-head.twig
```diff
@@ -15,9 +15,83 @@
...
+          "@context": "http://schema.org",
+          "@type": "WPHeader",
+          "@id": '"' + document.location.href + '#header' + '"'
+      }
+    </script>
+    <script type="application/ld+json">
+      {
+          "@context": "http://schema.org",
+          "@type": "WPFooter",
+          "@id": '"' + document.location.href + '#footer' + '"'
```
gender404: should be => "@id": document.location.href + '#footer'

styleguide/source/_meta/_00-head.twig
```diff
@@ -15,9 +15,83 @@
...
+  - header area
+  - footer area
+  - sidebar area
+  - main navigation area
+#}
+    <script type="application/ld+json">
+      {
+          "@context": "http://schema.org",
+          "@type": "WPHeader",
+          "@id": '"' + document.location.href + '#header' + '"'
```
gender404: should be => "@id": document.location.href + '#header'

gender404 (changes requested):
We're getting close.

ygannett:
I believe that the last change maintains JSON format with proper URLs.

styleguide/source/_meta/_00-head.twig
ygannett: Are you saying that document.location.origin is already wrapped with double quotations?

styleguide/source/_meta/_00-head.twig
ygannett: the target value is supposed to be used by Google in its search results page for the search box within a search result or a Information Graph card. So, it's nothing to do with twig itself.  That line should be output as: `"target":"https://www.mass.gov/?s={search_term_string}&post_type=post",`
If document.location.origin already contains double quotations at the beginning and the end, it needs to be changed back to 'https://www.mass.gov' since the whole thing needs to be one string.

styleguide/source/_meta/_00-head.twig
ygannett: If I understand your comments above,
`"logo": document.location.origin + '/{{ directory }}/images/stateseal-color.png',` outputs `"logo": "https://www.mass.gov"/PROPER_DIRECTORY/images/stateseal-color.png,`?  
Then, I'm afraid the logo location isn't recognized even {{ directory }} is replaced with a proper value.

gender404:
I hadn't seen this as being JSON format script tags.  That does change things.  I pushed a commit that should resolve the issues we're having.  Sorry about not noticing that earlier.

The main issue is that document.location doesn't work within a json type script tag.  It is just being outputted as text instead of as a variable.  What I found is that we can use JavaScript to create these json script tags.  Which means we can use dynamically generated data like document.location so that's what is in the commit I made.

If that commit doesn't work for schema.org, it can be reverted.

There is a section in here about it about a third of the way down:
https://moz.com/blog/using-google-tag-manager-to-dynamically-generate-schema-org-json-ld-tags

ygannett:
The URLs are output and formatted as expected:
```
    <script type="application/ld+json">
      { 
          "@context": "http://schema.org",
          "@type": "Organization",
          "url": "http://mayflower.mass.gov",
          "logo": "http://mayflower.mass.gov/assets/images/stateseal-color.png",
          "name": "Commonwealth of Massachusetts",
          "sameAs": [
            "https://www.facebook.com/massgov/",
            "https://twitter.com/massgov/",
            "https://www.linkedin.com/company/commonwealth-of-massachusetts/",
            "https://www.youtube.com/user/massgov/",
            "https://www.instagram.com/massgov/"
            ]
      }
    </script>
    <script type="application/ld+json">
      {
          "@context": "http://schema.org",
          "@type": "WPHeader",
          "@id": "http://mayflower.mass.gov/?p=pages-ACTION-get-a-state-park-pass#header"
      }
    </script>
    <script type="application/ld+json">
      {
          "@context": "http://schema.org",
          "@type": "WPFooter",
          "@id": "http://mayflower.mass.gov/?p=pages-ACTION-get-a-state-park-pass#footer"
      }
    </script>  
    <script type="application/ld+json">
        {
          "@context": "http://schema.org",
          "@type": "SiteNavigationElement",
          "@id": "http://mayflower.mass.gov/?p=pages-ACTION-get-a-state-park-pass#main-navigation"
        }
    </script>  
  </head>
```

styleguide/source/_meta/_00-head.twig
```diff
@@ -15,9 +15,82 @@
...
+  SITE LEVEL COMMON DATA: 
+  ORGANIZATION
+  - alternateName can be added later if helpful.
+#}
+    <script type="application/ld+json">
+      { 
+          "@context": "http://schema.org",
+          "@type": "Organization",
+          "url": "{{urlDomain}}",
+          "logo": "{{ urlDomain ~ '/' ~ directory ~ '/images/stateseal-color.png' }}",
```
gender404, line 53: A better pattern for this would be =>  "logo": "{{ urlDomain }}/{{ directory }}/images/stateseal-color.png",

This however does work so I'll merge the code in.

PR #305, opened 2017-01-19
Fixes DP-1312-act-paragraphs

powpow12:
In drupal, the link is not required but in the twig it was always show.  If an author didn't add a link the icon would show.  We add conditional logic to determine if the link has been populated, if so, display it.

To test:
 - go here `http://mass.local/bash-bish-falls-state-park`, confirm the the link icon isn't displayed (`>`)
- edit the page and add a link, confirm they are displayed correctly

Screenshots
screen shot 2017-01-19 at 3 16 07 pm
screen shot 2017-01-19 at 3 18 03 pm

### DIFF
--- a/styleguide/source/_patterns/02-molecules/action-activities.md
+++ b/styleguide/source/_patterns/02-molecules/action-activities.md
@@ -21,9 +21,15 @@ actionActivities: [{
     type: string/required
   description:
     type: string/required
+}]
+~~~
+
+### Required Variables
+~~~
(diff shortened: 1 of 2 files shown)


gender404:
This looks good, but could you please add this ticket to the release notes for Sprint 11.

https://github.com/massgov/mayflower/blob/dev/release-notes.md

Thanks.

powpow12:
done

PR #306, opened 2017-01-19
Release 3.1.0

jesconstantine:
### Org Page
* DP-708 - Implement Organization Page in PatternLab
* DP-867 - FE - Action Finder
* DP-874 - FE - Org: Implement Header

### Other
* DP-712 - Handle Stacked Row Optional Header Image variants
* DP-821 - Ensure that link chevron is never bumped to a new line

### _&& a bunch of other things from prior sprints to catch master up_


isaacchansky:
This seems good to me, though this is a lot to review.

PR #308, opened 2017-01-23
DP-1279: Add rich text atom for use with WYSIWYG editors.

gleroux02:
Ticket: https://jira.state.ma.us/browse/DP-1279

Creates an atom to display the "BLOB" from a rich text field. This would serve a need for any system that is implementing styleguide, but not for patternlab itself.


gender404 (changes requested):
I'm going to start over now that I've reviewed both Pull Requests.  It appears that you are doing something quite clever with the Rich-Text Pattern when used in Drupal by including a blob of html code instead of the array of elements I'm using in PatternLab.  

This seems like a great solution.

The naming of this atom as "rich-text" is, however, confusing since the content displayed isn't rendered to match a Rich Text editor.  I think it would be better to call it "html-embed" or something like that since it is just a raw html field.

gleroux02:
@legostud what about "raw-text" or "raw-html"?

gender404:
I can go with "raw-html"

gleroux02:
Yay, we named something! I'm sure we won't regret it in 5 minutes.

gender404:
probably.  I'm already stifling other potential names, like uncooked-html (raw).

gleroux02:
@legostud updated the name. And also removed the "Deprecated" state on paragraph as we can leave that alone.

styleguide/source/_patterns/01-atoms/11-text/raw-html.json
```diff
@@ -0,0 +1,5 @@
+{
+  "richText" : {
```
gender404: Sorry to be a stickler, but this should be rawHtml for consistency (should match the filename)

gender404 (changes requested):

gleroux02:
@legostud updated.

PR #309, opened 2017-01-24
Fixes Dp 1187 rte extra paragraph

powpow12:
Issue: Extra paragraph tag output on rendering of RTE paragraphs
https://jira.state.ma.us/browse/DP-1187

To test:
- Navigate to `http://mass.local/apply-nursing-license-new-graduate`
- Inspect 'Register for the NCLEX'
- confirm there is only one paragraph tag

Note: This twig template is called in several places:
image
As well as dynamically loaded in places like 
image 
which is rendered as the following
image


gender404:
This change doesn't work.  You've changed the paragraph.twig into, in essence, a RTE.

I think this is a larger issue than just the paragraph element.  I'm actually surprised that the Mayflower rich-text.twig file is being used in Drupal.  That organism isn't built like a true RTE.  It's built in a way that allows us to easily construct Rich Text for PatternLab without having to put html into our JSON data.  I would suggest overriding the rich-text organism with an actual RTE drupal element, making sure to keep the "ma__rich-text" and "js-ma-rich-text" classes to apply the correct styling.

powpow12:
this work is no longer needed, a better solution is being implemented here https://github.com/massgov/mayflower/pull/308

PR #313, opened 2017-01-26
Fixes DP-1521 Make callout links card a link

powpow12:
Issue:  Jsclickable js will always fire while interstitial js will not. This is only for links nested under jsclickable parents.  In html5, we can have block links - meaning divs, h1, p can
be children of the 'a' element. This removes the need for JS and will unblock this clash of JS for this element.

jira ticket: https://jira.state.ma.us/browse/DP-1521

Notes: html 5 resource, https://davidwalsh.name/html5-elements-links

### DIFF
--- a/release-notes.md
+++ b/release-notes.md
@@ -2,7 +2,7 @@
 
 ## 3.2.0 (Sprint 11)
 
-### DS-5 
+### DS-5
 * DP-1103 - Implement default file/download icon in Patternlab
 * DP-691 - Implement Google Translate in Pattern Lab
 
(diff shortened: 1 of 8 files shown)


gender404:
The CSS is currently looking for an a tag within the decorative link to apply styles.  that will need to be corrected

isaacchansky (approved):
👍

powpow12:
@legostud - sorry I missed this.  Can you be more specific about what css needs to be changed?

gender404:
@powpow12 - The main issue I saw was the SVG would wrap to a new line, because the structure and styling had changed to remove a needed layer of nesting.  I pushed up a change to fix the styling.

Also since we are no longer using the decorative link, I felt it would be better to just have everything within the callout-link namespace.  This way changes to the decorative link in the future will not causing issues with this component.

isaacchansky:
@legostud I'm jumping on this to try to finish it out - are there additional changes that need to be made here?

gender404 (approved):
This code looks code to me.

PR #314, opened 2017-01-26
DP-691 google translate

gender404:
Adding Google Translate to PatternLab

1. Removed Language Dropdown from utility nav
2. Added Google Translate code to the PatternLab foot.twig file
3. Added Google Translate html element to the Utility Nav
4. Updated various patterns to work with Translation code.


gleroux02 (changes requested):
The language dropdown does not appear to work when in a mobile viewport. I can get the language popup but after selecting a different language nothing happens. Note that I did not try this from a physical device.

gender404:
Well, I fixed what I believe is the issue.  I found that if I selected a language, select English, and then select another language Google Translate got very confused.  The bar at the top showed Arabic as the current language (regardless of what you picked) and everything was still in English.

Simple solution to get this ticket done was to remove English as one of the options.  I think this is going to end up being a bug with the Google Translate provided code.

PR #315, opened 2017-01-27
DP-1535 - Main Nav wide screen fixes

gender404:
This work is to allow touch screen devices to access the pull down menus

1. Added new "last-child" link to each drop down that links to the parent landing page
2. Added new Bent Arrow Icon
3. Converted top links into buttons

Other:
1. fixed gulp issue with SVG font files not getting copied to public
2. fixed font-family issue with form elements

Notes:
1. I did not update the Pilot version of the Main Nav since that Pattern should be obsolete now.


gleroux02:
The code looks fine and works fine. See my comment in https://jira.state.ma.us/browse/DP-1535 why I'm hesitant to merge.

PR #317, opened 2017-01-30
DP-981 header contacts

gender404:
This ticket is to update the Action Header pattern to use the Contact Us pattern instead of the custom Header Contact pattern.  Doing this will reduce duplication and help increase consistency.

1. Contact Group - made the title optional (including the icon)
2. Contact Us - Added --accordion modifier class to help style content 
3. Action Header
  * Added Contact Us pattern
  * left the Header Contact to support version 3.2
  * Moved the pageHeader data into the actionHeader namespace in a backward compatible manner
  * New CSS styling for laying out the Contact Us pattern.


gleroux02:
@legostud can you take a look at the conflicts?

gender404:
@gleroux02 - merge conflicts have been fixed

styleguide/source/_patterns/03-organisms/by-template/action-header.twig
```diff
@@ -1,7 +1,16 @@
...
+      {# for backward compatibility with version 3.2.0 #}
+      {% if actionHeader.pageHeader %} 
+        {% set pageHeader = actionHeader.pageHeader %}
+      {% endif %}
       {% include "@organisms/by-template/page-header.twig" %}
+      {% if actionHeader.contactUs %}
+        {% set contactUs = actionHeader.contactUs %}
+        {% include "@molecules/contact-us.twig" %}
+      {% endif %}
+      {# for backward compatibility with version 3.2.0 #}
```
gleroux02, line 13: Should this be an if else statement?

gender404: No.  The if statement is there to toggle the contact information on or off not to determine which pattern to use.

There is a ticket for Drupal to refactor the data for this pattern.  Once that is done, the headerContact if statement and contents will be removed.

styleguide/source/_patterns/02-molecules/contact-us.twig
```diff
@@ -18,7 +25,7 @@
   {% endfor %}
 
   {# Show 3rd+ contact groups inside of accordion (if they exist) #}
-  {% if contactUs.groups|length > 2 %}
+  {% if contactUsAccordion %}
```
gleroux02, line 28: Within drupal there is a field that let's the content editor set the display of the contact group. They can set it either full width, left or right. In this way they can set the address as full width. And then any subsequent groups as either left or right so that they display as two columns. I can't decide if it would be better to keep this approach and add a display variable here in patternlab so that each group is setting it's display. Or to remove this flexibility in drupal and have the order of the groups determine the display as you are here in the css. Do you have any thoughts on this? @legostud

gender404: I don't think we need to have that flexibility provided to the user.

Currently the first group, which should be the address, will always be full width.  The second group, if it is also the last group, will be full width.  If there are more than two groups, those get added to a right column and the second group reduces to 50% width.

gleroux02: I agree with you. Now it will just be a task of convincing the editors / removing that ability on the Drupal side.

PR #318, opened 2017-02-01
Dp 1759 org page map

gender404:
1. Added Optional title and link to Action Map
2. Added Action Map to the Org Page 
3. Removed a broken Action Section Map variant.
4. New modifier class added to single column layouts for styling purposes

Other:
1. Decorative Link - noticed we didn't have a variant in the dropdown for showing an external link


styleguide/source/assets/scss/02-molecules/_action-map.scss
```diff
@@ -1,12 +1,24 @@
...
-  @include ma-component-spacing;
-  height: 360px;
+  @include ma-component-padding;
 
-  @media ($bp-x-small-min) {
-    height: 500px;
+  .pre-content > & &__header,
+  .main-content--full .page-content > & &__header,
+  .post-content > & &__header {
+    @include ma-container;
```
gleroux02, line 7: @legostud when we get a chance to generalize this work I'd love to have a discussion about just using a container class so that we can get away from needing to write scss like this and just toss on one class that we need.

PR #327, opened 2017-02-06
DP-1759 - Org Page location row

gender404:
Request from Design QA to reduce the width of the location map row on the Org Page prompted me to rethink using the Action Map pattern.  I've decided a new Mapped Locations pattern would be better.

As part of this change, I also built a new Google Map pattern that will eventually replace the Action Map pattern.


styleguide/source/_patterns/02-molecules/google-map.md
```diff
@@ -0,0 +1,57 @@
+---
+title: Google Map
```
gleroux02: This extra character is causing the readme to fail.

gleroux02 (changes requested):
Do you know what that extra character is in the google-map.md? I'm unable to edit anything locally that let's me alter it and commit anything.

styleguide/source/_patterns/02-molecules/google-map.md
gender404: The only way I've fixed these to change the file encoding to unix.  Seems like an odd whitespace character

PR #331, opened 2017-02-09
Add audio file for video description below video

yriahi:
## Overview:
The PR adds a link to an audio files below the Youtube video in order to provide a description of what the video is about to meet accessibility needs. This references G159 specifically mentioned here https://www.w3.org/TR/WCAG20-TECHS/G159.html.

## Details:
Video-only content is inaccessible to people who are blind and to some who have low vision. Therefore, it is important for them to have an audio alternative. One way of doing this is to provide an audio track describing the information in the video. The audio should be a common audio format used on the internet, such as MP3. 

Example: 
A Web page has a link to a video-only presentation of a spaceship landing on Mars. The link to the video is a picture of a spaceship. Near the video is a link to an audio file of a person describing the video. This would look something like the following code example in HTML. 
```
Example Code: 
<a href="../video/marslanding.mp4"><img src="../images/spaceship.jpg" alt="Mars landing, video-only" width="193" height="255"/></a> 
<br /> 
<a href="Mars_landing_audio.mp3">Audio description of "Mars Landing"</a> 
```
Additional information on audio descriptions can be found here too: http://ncam.wgbh.org/invent_build/web_multimedia/accessible-digital-media-guide/adm-guidelines-appendices#ADM_Guidelines:_Appendix_5 , appendix 5 goes into detail in regard to creating audio descriptions.


gender404 (changes requested):
I'm not liking this solution.  

Technically, it meets the requirements but I think it fails the UX and design.  It does provide a link to an audio file, but since it isn't contained with the video area's styling it seems disconnected.   I think this should be reviewed by @reenybeeny to see if there is a better design for it.  Or if this is just for screen readers can we add a aria-hidden attribute? 

image

yriahi:
@legostud honestly, I am not really sure if this should be hidden or visual. I was following a recommendation provided to us by Perkins.org following an accessibility testing report. I would say we are open to either method that would enable the video component to pass accessibility testing.

gender404:
I think this needs more research.  From what I'm reading, if you provide a link to a video, you need to provide a link to an audio description as well.  What we have here is not a video link, but a video player.  So if feels as if we need to have an audio player instead of a link to an audio file.  Or we need a better video player that offers an audio only experience.

Maybe @isaacchansky might be able to help here.

isaacchansky:
@legostud yeah I'm with you - I think MVP might be an inline audio player, ideally a video solution that supports multiple audio tracks (one of which is the audio-description). 
The latter might be not possible since we're (I think) tied to youtube which doesn't support multiple audio tracks. 

But yeah, inline audio player w/ a label of "Audio Description for ____" probably makes sense. 

Not sure if this should only be for screen-readers. There may be individuals who have OK vision but movement is problematic so they avoid videos. If possible having it fit the design and be visible is ideal I think.

gender404:
Based on the conversation in the ticket, we're going to use a link, but it will take the user to a page with a text transcript of the video instead of using a link to an audio file.

I've pushed the code needed to accomplish this.  Please review and merge if it looks ok.

gender404 (approved):

styleguide/source/_patterns/01-atoms/09-media/video.twig
```diff
@@ -1,3 +1,9 @@
-<div class="ma__rich-text__embed js-ma-responsive-video">
-  <iframe width="853" height="480" src="{{video.src}}" frameborder="0" allowfullscreen title="{{video.title}}"></iframe>
+<div class="ma__video">
+  <div class="js-ma-responsive-video">
+    <iframe width="video.width" height="video.height" src="{{video.src}}" frameborder="0" allowfullscreen aria-label="video.label"></iframe>
+  </div>
+  {% if video.link %}
```
isaacchansky, line 5: Is this `.link` for the audio description? might be worth making it named more explicitly. Unless we're standardizing on a common 'link' json object, in which case maybe just a well-placed comment so it's clear what sort of link this is? (not to be confused with just an additional link to the video, a transcript, etc)

isaacchansky (approved):

PR #333, opened 2017-02-14
DP-1653 Map page

gender404:
Ticket DP-1653 - Create Map Page.

New:
1. Image Promo Molecule (twig, css)
2. Map Page (twig)
3. Location Listing Organism (twig, css, js)
4. Field Submit Molecule - simple text input with inline button
4. cssControlCode.js - checks if the css content for a given jQuery element is true or false
5. sticky.js - makes a given jQuery element sticky.

Updated
1. Link Atom - added aria-label field/value for screen readers
2. Decorative Link
  - Corrected Documentation
  - Added ability to show chevron as a third style variant
3. Action Activities
  - Gutted code and replaced with an include of the new image-promo 
4. Page Header - added new divider option (not actually used for this ticket)
5. Button Search - converted to use JSON data
6. Action Map JS - renamed to Google Map JS

Fixed
1. General Page - Data error with JSON values
2. Decorative Link - corrected documentation
3. RTE - JS wasn't adding the SVG icon to external links (wrong path)


styleguide/source/_patterns/02-molecules/image-promo.twig
```diff
@@ -0,0 +1,43 @@
...
+  <div class="ma__image-promo__details">      
+    <h2 class="ma__image-promo__title">
+      {% if imagePromo.title.href %}
+        {% set decorativeLink = imagePromo.title %}
+        {% include "@atoms/decorative-link.twig" %}
+      {% else %}
+        {{ imagePromo.title.text }}
+      {% endif %}
+    </h2>
+    {% if imagePromo.location %}
```
gleroux02, line 22: Should the location part of this be it's own molecule?

styleguide/source/_patterns/02-molecules/image-promo.twig
```diff
@@ -0,0 +1,43 @@
+<section class="ma__image-promo">
```
gleroux02, line 1: I'm looking to paint a bike shed... I'm not in love with the name "Image promo". Illustrated list makes more sense to me but what do you think? Also if we are adding other molecules this makes more sense to me as an organism.

styleguide/source/assets/js/modules/scrollAnchors.js
```diff
@@ -1,3 +1,5 @@
+import checkMobile from "../helpers/cssControlCode.js";
```
gleroux02, line 1: I'm unsure on how to test that this file has changed? What does it do and what should I be looking out for?

styleguide/source/assets/js/modules/zoomControls.js
```diff
@@ -1,28 +0,0 @@
-export default function (window,document,$,undefined) {
```
gleroux02, line 1: I'm unsure how to test this change.

styleguide/source/assets/js/modules/siteSettings.js
```diff
@@ -1,65 +0,0 @@
-import cookies from "../helpers/cookies.js";
```
gleroux02, line 1: I'm unsure how to test this change.

styleguide/source/_patterns/02-molecules/action-activities.twig
```diff
@@ -1,14 +1,32 @@
+{# DEPRECIATED - use image-promo instead #}
```
gleroux02, line 1: Should we update the action-activities.md to reflect that we have made this backwards compatible but it will still need to be refactored?

styleguide/source/_patterns/03-organisms/by-author/location-listing.json
```diff
@@ -0,0 +1,135 @@
...
+          "email": "mass.parks@state.ma.us",
+          "address": "24 Beacon St",
+          "address2": "",
+          "city": "Boston",
+          "state": "MA",
+          "zip": "02133"
+        }
+      }]
+    },
+    "results":[{
```
gleroux02: I'd like to use a generalized variable name here for the parent variable that holds all of the image-promo items. I think this should reflect the naming convention that is settled on for the image-promo molecule.

styleguide/source/_patterns/03-organisms/by-author/location-listing.json
```diff
@@ -0,0 +1,135 @@
+{
+  "locationListing": {
+    "form": {
```
gleroux02, line 3: Mainly a comment for myself but I want to look at this form further and make sure we are good on an implementation front.

styleguide/source/_patterns/02-molecules/image-promo.twig
gender404: problem is that it isn't a list.  It's just one thing, which is probably going to be used in a list, but could be used in other ways.  We've always defined Image, Title, Description, and Read more as an Image Promo.

Organism vs Molecule for me is defined by the html contained.  An organism should only include html for layout (divs mainly) so no links, images, or heading tags.

styleguide/source/_patterns/02-molecules/image-promo.twig
gender404: Are you thinking it might be used in other components?

styleguide/source/_patterns/03-organisms/by-author/location-listing.json
gender404: Good thought.  I was trying to think of the best way to implement a generic Form data structure.  For now, we just have the one one field-submit pattern, but the expanded version will contain several other fields.

styleguide/source/_patterns/03-organisms/by-author/location-listing.json
gender404: not sure I follow.  I can't really change results to imagePromo, because results is an Array and imagePromo is an object.

styleguide/source/assets/js/modules/scrollAnchors.js
gender404: These are the jump links on the Guide Page and Action Page.  The code changed is designed to change the JS behavior based on when the jump links visually change between narrow and wide screens.  On narrow screens, they act as a sticky drop down.  On wide screens the drop down is not used and the links stick to the left column of the Action Page.

I do see some sort of bug on Guide Page in narrow view.  I'll see if I can resolve that.

styleguide/source/assets/js/modules/siteSettings.js
gender404: you can tell if a js file is no longer needed by searching the patterns for the js class listed in the outer each statement.  "js-site-setting-form" in this case.  It's a JS pattern I used to detect if the item exists on a page and to allow for more than one of them.

gender404 (commented):
I think I answered all your questions.  let me know if I missed one.

styleguide/source/_patterns/02-molecules/image-promo.twig
gleroux02: What is catching my eye here is that the imagePromo is a pretty generalized style, or at least I think it should be, and the location is a very specific element being added in. If another component pops up that also requires a very specific element added to it then the template and data structure can get complicated in a hurry. I admit that I'm arbitrarily pointing out this specific case when we haven't done this anywhere else so I'm willing to say that we deal with this later.

styleguide/source/assets/js/modules/scrollAnchors.js
gleroux02: I'm seeing an error on http://localhost:3000/?p=pages-GUIDE-movng-to-ma-part1 when I click on the "In this guide" links. In any view mode. I'm taken to another page and get an error like ```Cannot GET /patterns/05-pages-GUIDE-movng-to-ma-part1/buying-or-renting-a-home```

styleguide/source/_patterns/03-organisms/by-author/location-listing.json
gleroux02: I don't think I was being clear. I'm wanting to be more consistent with our variable structure if possible so that we can start to write fewer functions on the system side. Currently we have many functions that do similar tasks but can't be generalized into 1 function because the output variable structure is slightly different in multiple places. 

What I'm trying to ask myself when reviewing new data structure is are we using the same variable names / array structure everywhere. So if you were including the imagePromo everywhere else would you be using results:[] as the array that contains them?

Again I'm arbitrarily pointing out this specific example when I would of just skipped over this in the past.

styleguide/source/assets/js/modules/siteSettings.js
gleroux02: Thank you, that is good to know. This looks fine then.

styleguide/source/assets/js/modules/zoomControls.js
gleroux02: There are no instances of classes using this JS so we should be fine here.

styleguide/source/_patterns/02-molecules/image-promo.twig
gleroux02: Good point. If you were to generalize the Action Activities into a general listing then "Action Activities" would be changed and the items it used would be an imagePromo. We've always called these things an illustrated list, and the individual element a illustrated list item. As I said I was starting a bike shed conversation, I'll concede and let's leave it alone.

gleroux02:
@legostud I have a couple further comments and one actual error.

nstriedinger:
On the infoWindow, we have our address stored as a textarea on the Drupal side.

nstriedinger:
@legostud Are you able to get the search to work with this google maps technique?

gender404:
@nstriedinger - The search isn't connected to anything.  It should be in a form that could reload the page with new data as an MVP solution.

The next version of this page includes more filters and pagination at which time we'll need to discuss who's responsible for refactoring the page to use a JS framework using an API for data, which has so far been outside my role.

styleguide/source/_patterns/03-organisms/by-author/location-listing.json
gender404: I can change results to imagePromos instead

styleguide/source/assets/js/modules/scrollAnchors.js
gender404: Somewhere along the line the JS code for Jump Links broke.  Should be working now.

gender404:
@gleroux02 - I think I've address all the concerns now.  Let me know if there are other things that need to change.

There are definitely larger questions around this page that we'll need to address in the future.  Currently, I'm working under an MVP mindset and just trying to get this page out the door with the minimum amount of work required so some things are not ideal (like the form section) and the new activeFilters message

The biggest question I have is how will this page function once it is more than just the one filter.  At that point, I would assume, we need to rethink using a twig template at all and instead use something like React JS to build the page dynamically pulling data from an API.

PR #337, opened 2017-02-16
DP-1670 ilustrated links

gender404:
Making the "Guide:" label optional.


gleroux02 (approved):

gleroux02:
@legostud http://localhost:3000/?p=pages-GUIDE-movng-to-ma-part2 doesn't have the "guide" label added to the json on the related guides. Is that on purpose or just missed?

PR #338, opened 2017-02-17
DP-1849 - Structured data - accessibility fixes (DS-8) 

ygannett:
Add missing context to the Directions links.
There is no visual difference to sighted users.  AT users get the additional information, a location name, after "Directions" in `<span>`.  The links are addressed as "Directions to LOCATION NAME."

The LOCATION NAME should match its contact name.

Set up necessary a variable and its value to pass on to the `/_patterns/01-atoms/decorative-link.twig` for `<span class="ma__decorative-link__info">` in `/_patterns/02-molcules/contact-group.twig`.

```
<span class="ma__decorative-link">
      <a href="https://www.google.com/maps/place/555+Park+Dr,+Boston,+MA+02215" class="js-clickable-link" property="url">Directions<span class="ma__decorative-link__info"> to LOCATION NAME</span>&nbsp;<svg aria-hidden="true" id="SvgjsSvg1000" xmlns="http://www.w3.org/2000/svg" version="1.1" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:svgjs="http://svgjs.com/svgjs" width="17" height="16" viewBox="0 0 17 16"><defs id="SvgjsDefs1001"/><path id="SvgjsPath1007" d="M739.199 92.7986L739.199 80L735.9989999999999 80L735.9989999999999 95.9983L751.9979999999999 95.9983L751.9979999999999 92.7986ZM743.294 81.0032L743.305 83.5605L746.6049999999999 83.5737L742.9939999999999 87.1849L744.809 89.0001L748.4259999999999 85.3836L748.4399999999999 88.6956L750.997 88.7058L750.976 83.59129999999999L750.966 81.03389999999999L748.409 81.02349999999998L748.409 81.02369999999999Z " transform="matrix(1,0,0,1,-735,-80)"/></svg>
</a>
  </span>
```


gender404 (changes requested):
You can't use contactUs within contactGroup.  if contactGroup was to be used in a different organism say headerContact then contactUs wouldn't be valid or worse it could get that value from a contactUs organism placed on the page above it.

ygannett:
Thank you so much for reviewing this.  I'm stuck now.
What is the best way to obtain the LOCATION NAME for the contact group in this case?

gender404:
We've pushed the data model for the contact group as far as it could.  It was time for an update to pull out the link data into it's own object so we can properly pass in "info" and "property" values.  I made these change in a way that should be backward compatible with the old data model.

To make this work in Drupal, they will need to update there data object to match this new one.

gender404 (approved):

PR #339, opened 2017-02-17
DP-1995 - Add nl2br filter to contact us address to allow new line instead of <br>

gleroux02:
Ticket: https://jira.state.ma.us/browse/DP-1995

Uses http://twig.sensiolabs.org/doc/2.x/filters/nl2br.html on the address in the contact group to allow the use of /n for a line break.

To test:
* Confirm that an address that is supposed to have a multi line break uses \n instead and that it renders properly.

### DIFF
--- a/styleguide/source/_patterns/02-molecules/action-map.json
+++ b/styleguide/source/_patterns/02-molecules/action-map.json
@@ -30,7 +30,7 @@
         "phone": "16176261250",
         "fax": "16176261351",
         "email": "mass.parks@state.ma.us",
-        "address": "251 Causeway Street, Suite 900<br />Boston, MA 02114-2104"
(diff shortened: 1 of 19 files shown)


gender404 (changes requested):
You either missed a file or you accidentally changed an address that should have kept the ```<br />```.

The google-map.twig file has an address that is displayed in the map's Info Window.  I'm not sure if the address for Info Window should keep the ```<br />``` or be replaced with "\n" instead.

gender404 (approved):
I uploaded changes for the google map to match the same format

gleroux02:
@legostud This leaves a <br> displaying in the pop up. Is google just looking to make it one line? If that is the case why don't we replace "/n" with just a space?
image

gender404:
Your screen shot looks out of date.  Are you using the updated handlebar template?

image

PR #340, opened 2017-02-20
Organism callout links

gleroux02:
Adds the ability to add more then one callout link at once.


gender404:
FYI, this Organism already existed => "Key Actions"

PR #348, opened 2017-02-23
Location Listing (MVP) Design QA

reenybeeny:
- wrapped directions link in a new class and styled the text to be smaller
- adjusted font size of address and added margin
- removed underline on address
- added a negative bottom margin on the map so there isn’t a gap between it and the footer
- adjusted font, border, line-height, color, etc. on Results header
- added input placeholder styles to all form inputs to meet accessibility
- updated filter button to gray color
- increased line-height on map info window text

### DIFF
--- a/styleguide/source/assets/scss/02-molecules/_image-promo.scss
+++ b/styleguide/source/assets/scss/02-molecules/_image-promo.scss
@@ -42,10 +42,7 @@ $image-promo-bp-x-large-min: "min-width: 1160px";
 
   &__description {
     margin-top: 10px;
-
-    p {
-      font-size: 1.25rem;
-    }
+    font-size: 1.25rem;
   }
 
   &__location {
--- a/styleguide/source/assets/scss/02-molecules/_image-promo.scss
+++ b/styleguide/source/assets/scss/02-molecules/_image-promo.scss
@@ -41,16 +41,15 @@ $image-promo-bp-x-large-min: "min-width: 1160px";
   }
 
   &__description {
+    margin-top: 10px;
 
-  }
-
-  &__link {
-    @include ma-chevron;
-    @include ma-link-underline;
-    padding-bottom: 3px;
+    p {
+      font-size: 1.25rem;
+    }
   }
 
   &__location {
+    font-size: 18px;
     padding: 5px 0;
 
     svg {
@@ -63,7 +62,6 @@ $image-promo-bp-x-large-min: "min-width: 1160px";
 
     a {
       @include ma-link-underline;
-      border-bottom-width: 1px;
       display: inline;
     }
   }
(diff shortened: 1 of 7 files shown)


styleguide/source/assets/scss/02-molecules/_image-promo.scss
```diff
@@ -41,16 +41,15 @@ $image-promo-bp-x-large-min: "min-width: 1160px";
...
 
-  }
-
-  &__link {
-    @include ma-chevron;
-    @include ma-link-underline;
-    padding-bottom: 3px;
+    p {
+      font-size: 1.25rem;
+    }
```
gender404: You should add this font-size to the &__description instead.  The description is a Rich Text editor so you could have various other elements used in there that should all have a reduced font-size.

styleguide/source/assets/scss/02-molecules/_image-promo.scss
```diff
@@ -41,16 +41,15 @@ $image-promo-bp-x-large-min: "min-width: 1160px";
...
   }
 
   &__location {
+    font-size: 18px;
     padding: 5px 0;
```
gender404, line 50: not rems?  1.125rem

styleguide/source/assets/scss/03-organisms/_location-listing.scss
```diff
@@ -41,6 +41,7 @@
     @media ($bp-large-min) {
       height: 100vh;
       max-height: none;
+      margin-bottom: -60px;
```
gender404: This isn't great, but I can't think of a better solution.  This is, however, on the wrong element.  putting it here breaks the sticky map JavaScript.  This should be moved to the &__container instead.  up to you if you want to add some extra padding back to the results column

styleguide/source/assets/scss/06-theme/02-molecules/_location-listing.scss
```diff
@@ -1,6 +1,25 @@
...
+      }
+    }
   }
-}
\ No newline at end of file
+
+  &__results {
+
+    & > p {
+      @include ma-font(700);
```
gender404: this mixin isn't needed anymore.  just use font-weight: 700 instead

gender404 (changes requested):

styleguide/source/assets/scss/02-molecules/_image-promo.scss
reenybeeny: I did this since I didn't want it to change size on breakpoints. You mentioned it was ok to do that on elements that would have the same size no matter what. Is that still ok?

gender404: that's fine then.  I just wanted to make sure

PR #354, opened 2017-03-01
Force pageBanner.titleSubText to uppercase and put in ( )'s

jesconstantine:
This PR adds some formatting to the pageBanner.titleSubText property.  Namely, it forces the string to uppercase, and surrounds it with ().  I'm certainly open to making this behavior conditional and based on a boolean flag (e.g. pageBanner.titleSubText.isAcronym), but since the .json data sent for this page template was (EOHHS) I figured the design decision was already made.  Let me know if that's not the case.

Related to JIRA ticket: https://jira.state.ma.us/browse/DP-2008

To test:

- [ ] Pull the branch down and build locally
- [ ] Browse to the org page: http://localhost:3000/?p=pages-ORG-Health-Services
- [ ] Confirm that the subtitle text still renders in all capital letters and in parens.
- [ ] Optionally, alter the pageBanner.subTitleText value in mayflower patterns pages > ORG-Health-Services.json

### DIFF
--- a/styleguide/source/_patterns/03-organisms/by-template/page-banner.twig
+++ b/styleguide/source/_patterns/03-organisms/by-template/page-banner.twig
@@ -27,7 +27,7 @@
       <h1 class="ma__page-banner__title">
         {{ pageBanner.title }}
         {% if pageBanner.titleSubText %}
-          <span> {{ pageBanner.titleSubText }}</span>
+          <span> ({{ pageBanner.titleSubText|upper }})</span>
         {% endif %}
       </h1>
     </div>
--- a/styleguide/source/_patterns/05-pages/ORG-Health-Services.json
+++ b/styleguide/source/_patterns/05-pages/ORG-Health-Services.json
@@ -16,7 +16,7 @@
     "size": "large",
     "icon": null,
     "title": "Executive Office of Health and Human Services",
-    "titleSubText": "(EOHHS)"
+    "titleSubText": "eohhs"
   },
 
   "actionHeader": {


jesconstantine:
I wonder if we shouldn't also update the name of this page .twig template to be more general?  If we use these page level templates in Drupal, it will be odd to reference ORG-Health-Services, etc.

styleguide/source/_patterns/03-organisms/by-template/page-banner.twig
```diff
@@ -27,7 +27,7 @@
       <h1 class="ma__page-banner__title">
         {{ pageBanner.title }}
         {% if pageBanner.titleSubText %}
-          <span> {{ pageBanner.titleSubText }}</span>
+          <span> ({{ pageBanner.titleSubText|upper }})</span>
```
lukewertz, line 30: Shouldn't this be `&nbsp;<span>({{...` rather than `_<span>({{`. Seems odd to be to be hard-coding whitespace in this way.

gender404: The thought process behind using (EOHHS) as content was to provide content editors with more flexibility.  There is no reason this field has to always be an abbreviation.  The page-banner pattern is also used on the Topic and Subtopic pages so they could use it there in a different way.  

As it is currently being used, I would assume the uppercase and "( ... )" logic would be part of the content type rules and not part of the view's logic.

And I would agree with Luke, that I should have used &nbsp; instead of _

gender404:
I would seriously hope you never use the Page patterns in Drupal.  Those are created just for getting designs approved and as examples of how a page could be constructed.

When we start doing Content Types, new templates should be built for them.  Those templates would be a better pattern to include since they will have a proper definition of what components can be placed where instead of just using four main placeholders.

We can then update these Pages to use the proper content type templates to show real content.

styleguide/source/_patterns/03-organisms/by-template/page-banner.twig
lukewertz: @legostud Don't think of this as a field in the CMS, though... Think of this as "the display of this information in this context." With the info in a field in the CMS, we may want to re-use it (like on the 'View all services by EOHHH ->' link).

I'd be okay saying we don't upper-case it, but do add the paren in the template.

jesconstantine: Does it make sense to check for the existence of flags and conditionally apply the ( ) and uppercase filter in this twig template?

gender404: @lukewertz - Is there a way to wrap the alternative name with ( ) when this content type is used?  Maybe using a view mode?

jesconstantine: @legostud this is kind of what I was getting at with my last comment above, except I think Drupal (the content type) should only be responsible for setting the flag.  I think the logic and rendering should happen in Mayflower so that other systems can take advantage of this functionality.  And who knows where else we'll use the pattern in the future.  What do you think?

gender404: It still gets tricky even if we use a flag.  Simple flag would be a boolean => parenthesis: true, but what if another system or content type wants to wrap it in quotes for example.  Then you create a type variable that equals "parenthesis" to add the () version and then create an if else for the quote version which then grows into a switch statement.

gender404:
Any objections if I close this Pull Request?

PR #357, opened 2017-03-06
DP-1370 Radio buttons and checkboxes more visible in feedback forms

mrossi113:
Accessibility fix to make the radio buttons and check boxes more visible on feedback form. The border color for these were changed to #757575 per Sarena and Harlan request in the comments. These checkbox and radio buttons borders were changed from border-color: $c-bd-input; to border-color: #757575;

JIRA ticket: https://jira.state.ma.us/browse/DP-1370

To test:
Compare the radio button with the current pilot.mass.gov/topics/health-social-services feedback form.
Inspect the radio button and change the .ma__input-radio border color to #757575
To compare the check box go to http://localhost:3000/?p=pages-pilot-transition vs current site https://pilot.mass.gov/leaving-pilot?continueURL=http://www.mass.gov/dor/child-support/court-information/paternity-establishment/


gender404 (commented):
We have been trying to not use colors directly in SASS files to help maintain consistency.  To make this #757575 you should either create a new variable $c-bd-input-dark or change the value of $c-bd-input to #757575.

Variables are named using a pattern:
c => color variable
bd => applies to borders
input => descriptive name
dark => optional modifier

I'm going to go out on a limb here and say that we are most likely going to need to change the border color on all inputs to #757575 for readability.  

You just need to search the scss file to determine what other items are using this variable before updating.

reenybeeny:
@legostud / @mrossi113 - it was decided not to change the border color on inputs or dropdowns at this time, so the change should only apply to radio buttons and checkboxes.

mrossi113:
@legostud / @reenybeeny  Okay I will create a new variable in the SASS file with #757575 for just these two radio buttons and checkboxes border color.

isaacchansky:
Probably a good idea to darken the borders, but the real purpose of this ticket is to fix the issue of identifying _focus_.

So the user-story here is "as someone with potential visual impairment, it should be easy for me to tell where my keyboard focus is." 

It looks like the focus styles haven't changed, they're still:
```
outline-color: -webkit-focus-ring-color;
```

gender404:
The radio button is missing the outline.  You can try adding that to make it more visible.

.ma__input-radio input:focus ~ label:before {
  outline: -webkit-focus-ring-color auto 5px;
}

.ma__input-radio label:before {
 ... 
 overflow: hidden;
  ...
}

styleguide/source/assets/scss/06-theme/00-base/_colors.scss
```diff
@@ -34,6 +35,6 @@ $c-bg-section: $c-gray-f2f2f2;
 
 $c-bd-divider: $c-gray-dcdcdc;
 $c-bd-input: $c-gray-dcdcdc;
-
+$c-bd-input-dark: $c-gray-757575;
```
gender404, line 38: nicely done

styleguide/source/assets/scss/01-atoms/_forms.scss
```diff
@@ -75,6 +76,7 @@ label {
         left: 0;
       text-align: center;
       width: 2em;
+      overflow: hidden;
```
gender404, line 79: I don't believe this is needed here.  I added it to the radio button due to the font-size being larger than the height.  

If you want to see an interesting outline shape, use the inspector tools to turn overflow:hidden off on the radio button before element.

styleguide/source/assets/scss/06-theme/01-atoms/_forms.scss
```diff
@@ -106,6 +106,7 @@ label {
      */
     @media (-webkit-min-device-pixel-ratio:0) {
       outline-color: -webkit-focus-ring-color;
+      outline: -webkit-focus-ring-color auto 5px;
```
gender404, line 109: outline includes outline-color so you don't need both.

styleguide/source/assets/scss/06-theme/01-atoms/_forms.scss
```diff
@@ -132,6 +133,7 @@ label {
...
```
gender404, line 136: outline includes outline-color so you don't need both.

mrossi113:
Talked with @WheresHJ about the border color and the focus for checkboxes and radio buttons. From the discussion was to keep the #757575 as the border color as well as the suggestion from @legostud 

`.ma__input-radio input:focus ~ label:before {
outline: -webkit-focus-ring-color auto 5px;
}
.ma__input-radio label:before {
...
overflow: hidden;
...
}`

@isaacchansky Harlan said he would talk with you about these changes.

gender404 (changes requested):
Just a few changes

isaacchansky:
Cool this looks good for now. We'll plan to circle back later and look at all focus states holistically to ensure that we're consistent and deliberate about how the whole system works together.

styleguide/source/assets/scss/01-atoms/_forms.scss
mrossi113: Yes the `overflow:hidden` was not needed for the checkboxes only the radio buttons. Made the change.

styleguide/source/assets/scss/06-theme/01-atoms/_forms.scss
mrossi113: Removed the outline-color

styleguide/source/assets/scss/06-theme/01-atoms/_forms.scss
mrossi113: Removed the outline-color

isaacchansky:
Sorry, closed this too early: adjustments @legostud suggested are being done here: https://github.com/massgov/mayflower/pull/365/files

PR #361, opened 2017-03-07
DP-1893 [a11y] add the status of alert message display with aria

ygannett:
No visual change.

Add `aria-expanded` to the alert `<button>` to tell the current status of the display of the alert messages. Its value gets changed between `true` and `false` with JavaScript as the button is clicked.

## Testing
- [ ] The `aria-expanded` value is `false` when the alert messages are hidden.
button_close
- [ ] The `aria-expanded` value is `true` when the alert messages are shown.
button_open
- [ ] The status gets updated as the `<button>` is clicked accordingly.

### DIFF
--- a/styleguide/source/_patterns/02-molecules/button-alert.twig
+++ b/styleguide/source/_patterns/02-molecules/button-alert.twig
@@ -1,4 +1,4 @@
-<button class="ma__button-alert" aria-expanded="false">
+<button class="ma__button-alert">
   <span class="ma__button-alert__hide">{{buttonAlert.hideText}}</span>
   <span class="ma__button-alert__show">{{buttonAlert.showText}}</span> {{buttonAlert.text}}
 </button>
--- a/styleguide/source/_patterns/02-molecules/button-alert.twig
+++ b/styleguide/source/_patterns/02-molecules/button-alert.twig
@@ -1,4 +1,4 @@
-<button class="ma__button-alert">
+<button class="ma__button-alert" type="button">
   <span class="ma__button-alert__hide">{{buttonAlert.hideText}}</span>
   <span class="ma__button-alert__show">{{buttonAlert.showText}}</span> {{buttonAlert.text}}
 </button>
--- a/styleguide/source/_patterns/02-molecules/button-alert.twig
+++ b/styleguide/source/_patterns/02-molecules/button-alert.twig
@@ -1,4 +1,4 @@
-<button class="ma__button-alert">
+<button class="ma__button-alert" aria-expanded="false">
   <span class="ma__button-alert__hide">{{buttonAlert.hideText}}</span>
   <span class="ma__button-alert__show">{{buttonAlert.showText}}</span> {{buttonAlert.text}}
 </button>
--- a/styleguide/source/assets/js/modules/emergencyAlerts.js
+++ b/styleguide/source/assets/js/modules/emergencyAlerts.js
@@ -7,7 +7,8 @@ export default function (window,document,$,undefined) {
         open = true,
         id = $el.data('id'),
         cookieName = 'emergency-alerts' + id,
-        cookieValue = cookie.getCookie(cookieName);
+        cookieValue = cookie.getCookie(cookieName),
+        $buttonState = $(this).find('.ma__button-alert').attr('aria-expanded');
 
     if(typeof(cookieValue) != 'undefined' && cookieValue === 'false') {
       // cookieValue is a string so we can't use the value directly
@@ -21,9 +22,12 @@ export default function (window,document,$,undefined) {
     $el.on('click','.js-accordion-link',function(){
       // toggle the current state
       open = !open;
-      // update open/close state cookie 
+      // update open/close state cookie
       // leave off third argument to make it expire on session
       cookie.setCookie(cookieName,open);
+      // change the state of aria-expanded
+     $buttonState === 'false' ? $buttonState = 'true' : $buttonState = 'false';
+     $(this).find('.ma__button-alert').attr('aria-expanded',$buttonState);
     });
 
   });
--- a/styleguide/source/assets/js/modules/emergencyAlerts.js
+++ b/styleguide/source/assets/js/modules/emergencyAlerts.js
@@ -1,6 +1,10 @@
 import cookie   from "../helpers/cookies.js";
 
 export default function (window,document,$,undefined) {
+  // Emergency Alerts start close on page load
+  // the default behavior is to expand the alerts
+  // Emergency Alerts should stay closed if the cookie is set to false
+
   $('.js-emergency-alerts').each(function(){
     let $el = $(this),
         open = true,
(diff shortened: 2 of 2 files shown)


styleguide/source/assets/js/modules/emergencyAlerts.js
```diff
@@ -21,11 +22,14 @@ export default function (window,document,$,undefined) {
     $el.on('click','.js-accordion-link',function(){
       // toggle the current state
       open = !open;
-      // update open/close state cookie 
+      // update open/close state cookie
       // leave off third argument to make it expire on session
       cookie.setCookie(cookieName,open);
+      // change the state of aria-expanded
+     $buttonState === 'false' ? $buttonState = 'true' : $buttonState = 'false';
```
gender404: for future reference this would be written $buttonState = $buttonState === 'false' ? 'true' : 'false';

styleguide/source/assets/js/modules/emergencyAlerts.js
```diff
@@ -7,7 +7,8 @@ export default function (window,document,$,undefined) {
         open = true,
         id = $el.data('id'),
         cookieName = 'emergency-alerts' + id,
-        cookieValue = cookie.getCookie(cookieName);
+        cookieValue = cookie.getCookie(cookieName),
+        $buttonState = $(this).find('.ma__button-alert').attr('aria-expanded');
```
gender404: This isn't the right approach.  
1. You're using a CSS class to target an element using JS.  You should only use JS classes ('.js-...')
2. This button may not always be an accordion button so having the attribute in the code is not ideal.  Would be better to have the JS add the attribute instead.
3. We already have a variable (open) tracking the state of the accordion so we can just use that variable instead.

styleguide/source/assets/js/modules/emergencyAlerts.js
```diff
@@ -21,11 +22,14 @@ export default function (window,document,$,undefined) {
     $el.on('click','.js-accordion-link',function(){
       // toggle the current state
       open = !open;
-      // update open/close state cookie 
+      // update open/close state cookie
       // leave off third argument to make it expire on session
       cookie.setCookie(cookieName,open);
+      // change the state of aria-expanded
+     $buttonState === 'false' ? $buttonState = 'true' : $buttonState = 'false';
+     $(this).find('.ma__button-alert').attr('aria-expanded',$buttonState);
```
gender404: This made me realized that the click event should have been on the button not the div wrapping it.

Once that change is made, this can be simplified to $el.attr('aria-expanded', open);

I'm pretty certain we can pass a boolean value to the attribute, but if it doesn't work then use a ternary to switch it to a string value => open ? 'true' : 'false';

styleguide/source/assets/js/modules/emergencyAlerts.js
```diff
@@ -21,11 +22,14 @@ export default function (window,document,$,undefined) {
     $el.on('click','.js-accordion-link',function(){
```
gender404: There appears to be an existing issue here.  The click really should be on the button not the js class and we're missing the e.preventDefault(); to make sure we don't accidentally submit a form on the page.

$button.trigger('click');

and 

$button.on('click', function(e) {
  e.preventDefault();

styleguide/source/assets/js/modules/emergencyAlerts.js
```diff
@@ -7,7 +7,8 @@ export default function (window,document,$,undefined) {
         open = true,
         id = $el.data('id'),
         cookieName = 'emergency-alerts' + id,
-        cookieValue = cookie.getCookie(cookieName);
+        cookieValue = cookie.getCookie(cookieName),
+        $buttonState = $(this).find('.ma__button-alert').attr('aria-expanded');
 
     if(typeof(cookieValue) != 'undefined' && cookieValue === 'false') {
       // cookieValue is a string so we can't use the value directly
```
gender404: in this conditional we should also include an update to the button.
$button.attr('aria-expanded', open);

styleguide/source/assets/js/modules/emergencyAlerts.js
gender404: I would suggest we create a new variable to save the button element.
$button = $el.find('.js-accordion-link button');

styleguide/source/_patterns/02-molecules/button-alert.twig
```diff
@@ -1,4 +1,4 @@
-<button class="ma__button-alert">
+<button class="ma__button-alert" aria-expanded="false">
```
gender404: Let's leave this off since the button could be used elsewhere and not need aria-expanded.  We'll have the JS add it instead.

gender404 (changes requested):
Sorry for all the small changes.  They kind of built on top of each other as I was reviewing.

styleguide/source/assets/js/modules/emergencyAlerts.js
ygannett: Would you like to have everything in `$el.on('click','.js-accordion-link',function(){}` to be in `$button.on('click', function(e) {}`, or just the what I added?  Just making sure, I don't know how the cookie stuff would be affected by this.

gender404: everything should be in there.

styleguide/source/assets/js/modules/emergencyAlerts.js
```diff
@@ -1,31 +1,42 @@
...
   $('.js-emergency-alerts').each(function(){
     let $el = $(this),
         open = true,
         id = $el.data('id'),
         cookieName = 'emergency-alerts' + id,
-        cookieValue = cookie.getCookie(cookieName);
+        cookieValue = cookie.getCookie(cookieName),
+        $button = $el.find('.js-accordion-link button'),
+        // Add an aria-expanded attribute with its default value, false. to the <button>
+        $buttonState = $button.attr('aria-expanded', !open);
```
gender404: $buttonState = $button.attr('aria-expanded', !open);
This line isn't needed.  It will be set either by line 18 or by line 37 to the proper value.

styleguide/source/assets/js/modules/emergencyAlerts.js
ygannett: Ok.  Thank you.

styleguide/source/_patterns/02-molecules/button-alert.twig
ygannett: That has been removed from this template and inserted by JS.

styleguide/source/assets/js/modules/emergencyAlerts.js
ygannett: Used the open variable to all related areas.

styleguide/source/assets/js/modules/emergencyAlerts.js
```diff
@@ -1,31 +1,42 @@
...
       // toggle the current state
       open = !open;
-      // update open/close state cookie 
+      // update open/close state cookie
       // leave off third argument to make it expire on session
       cookie.setCookie(cookieName,open);
+      // change the state of aria-expanded
+      $buttonState = $buttonState === open ? !open : open;
+      $(this).attr('aria-expanded', $buttonState);
     });
```
gender404: This can be reduced to just $(this).attr('aria-expanded', open);

styleguide/source/assets/js/modules/emergencyAlerts.js
```diff
@@ -1,31 +1,42 @@
...
+    $button.on('click', function(e) {
+      e.preventDefault();
       // toggle the current state
       open = !open;
-      // update open/close state cookie 
+      // update open/close state cookie
       // leave off third argument to make it expire on session
       cookie.setCookie(cookieName,open);
+      // change the state of aria-expanded
+      $buttonState = $buttonState === open ? !open : open;
```
gender404: This isn't needed and doesn't make any sense to me.  Sorry to be blunt, but that is a very odd line of code.  You are basically saying if the two values are equal make them not match and if they don't match make then equal.

gender404 (changes requested):
This is much better, but there are still a couple of changes.

ygannett:
I noticed some inconsistent behaviors when the page is loaded. Sometimes the `<button>` has `aria-expanded="false"`, sometimes it doesn't.  It seems that it doesn't have the attribute when the page is loaded for the first time (= no cache, no history of the alert).
alerts-first_load

When the button is clicked with this condition, the attribute's value is set opposite.  It's set to `true` when it's supposed to be `false`.

@isaacchansky helped me to see it's coming from the condition;
```
 if(typeof(cookieValue) != 'undefined' && cookieValue === 'false') {
      // cookieValue is a string so we can't use the value directly
      open = false;

      $button.attr('aria-expanded', open);
    }
```
When the condition is changed to
```
if(typeof(cookieValue) === 'undefined' || cookieValue === 'false')
```
the attribute is inserted to the `<button>` every time the page is loaded with the correct value, and the value is updated correctly with clicks.

So, the question is that how this change affects the cookie. I don't know what's this cookie is about. Should the cookie and the aria-expanded have their own conditions for each to maintain their own features?

gender404:
Thank you for finding that bug.  It was interesting to figure out what was actually happening in the code.

Originally, I was triggering the click on the js-accordion-link element and everything appeared to be working fine.  In reality it wasn't quite doing what I thought.  I had assumed that it was calling the click event in emergencyAlerts, but since the click event hadn't been bound yet, it was just calling the accordion click event and nothing else.  This gave the impression that things were working properly.

It also appears that we were triggering the accordion twice since we have two buttons in the code now so I'm guessing that also added some issues.

I pushed up some changes and it should be working fine now.

PR #366, opened 2017-03-13
DP-2181 - Updated Guide Pages

gender404:
1. Combined the three Guide Pages into one page (based on invision design)
2. Alert Callout
  - Made the href value optional
  - Modified the twig to wrap the content in an anchor tag
3. Updated Rich Text and Action Steps to use a new 'restricted' container width mixin
  - sets a max-width of 820px to the content contain for better readability
  - based on what use to be in the Rich Text editor CSS, but is now more re-usable


styleguide/source/_patterns/03-organisms/by-author/callout-alert.twig
```diff
@@ -1,9 +1,30 @@
...
-  <div class="ma__callout-alert__icon">
-    {% include "@atoms/05-icons/svg-alert.twig" %}
-  </div>
-  <div class="ma__callout-alert__link">
-    {% set decorativeLink = calloutAlert.decorativeLink %}
-    {% include "@atoms/decorative-link.twig" %}
-  </div>
-</section>
\ No newline at end of file
+  {% if ca.href %}
```
gleroux02: @legostud are there other instances where we are shortening variables in this manner? Specifically calloutAlert to ca.

gender404: this is the only place.  It was getting tiresome to write calloutAlert.decorativeLink... everytime I need to access a variable.  Is this a bad practice?  I can change ca to calloutAlert instead.

PR #367, opened 2017-03-14
DP-2181 Guide page - RTE Icon List

gender404:
New icon list pattern for Rich Text editor


styleguide/source/_patterns/01-atoms/06-rich-text/icon-list.twig
```diff
@@ -0,0 +1,12 @@
+<ul class="ma__icon-list">
```
gleroux02, line 1: @legostud can you resolve the tab spacing on this file? Oddly I cannot solve this myself, even directly editing through github.

gender404: Fixed.  Looks like my editor decided to use tabs instead of spaces for some reason.

PR #368, opened 2017-03-14
Dp 2207 org landing page content type

jesconstantine:
This PR creates a content type template for the org landing page.

JIRA Ticket: https://jira.state.ma.us/browse/DP-2207


jesconstantine:
I don't think we're quite ready for a content type template yet.  Closing for now.

PR #373, opened 2017-03-17
Dp 2114 service page

gender404:
This is still being worked on, but I'm hoping to wrap it up today.

Ticket:
https://jira.state.ma.us/browse/DP-2114

Goal - Create new Services Page.

New Patterns:
- Teaser Text - Rich Text with an optional decorative link (learn more)
- Services Template - Based on stacked row, but with some additional classes

Updated Patterns:
- Video - added float right option
  - Added new data variable => position
- Action Finder - New view triggered by removing the background image
  - Added new variables => Id and hideFilter
- Page Banner - New overlay style 
  - Added new variable => color
  - Added new value to size => 'overlay'
- Link list - Added controls for heading type and single vs double column
  - Added new variable => stacked to trigger the single column view
  - changed variable 'title' to be either compHeading or sidebarHeading
  - added v3.5 code to set compHeading if the 'title' is still being used
- Split Columns - adjusted the spacing for 50/50 split.
- Mapped Locations
  - new variables => info and property


jesconstantine:
@legostud  Do you know if this PR is touching so many files because we never synced master back down to dev after the release?  cc: @isaacchansky

isaacchansky:
Doesn't seeeeem to be related - the attached img is the visualized state of git.

screen shot 2017-03-17 at 12 33 10 pm

gender404:
@lukewertz - Services page

gender404:
@jesconstantine, @isaacchansky - Yes, the extra files in there are because dev doesn't include master.  I always create my branches off of master to make sure I'm working with clean code to start.  This process works well with small changes.

This, being a larger task, required a merge from dev so the code isn't quiet as clean as I prefer.

styleguide/source/_patterns/03-organisms/by-author/link-list.twig
```diff
@@ -1,24 +1,49 @@
 <section class="ma__link-list">
   <div class="ma__link-list__container">
+    {# added for backward compatibilty with v3.5 #}
```
gleroux02: The changes to the link list seem to be an issue on the org page. 
screen shot 2017-03-20 at 4 13 59 pm

styleguide/source/_patterns/04-templates/01-content-types/services.twig
```diff
@@ -0,0 +1,80 @@
+{% include "@organisms/by-template/header.twig" %}
```
gleroux02: Should the services here also show the "Related Services" and "Additional Resources" link lists that display in the 05-pages directory? I guess I'm not sure why we need to have it in both places.

styleguide/source/_patterns/04-templates/01-content-types/services.twig
```diff
@@ -0,0 +1,80 @@
...
+    {% endblock %}
+  </div>
+  <div class="stacked-row-content">
+    <section class="ma__stacked-row-section">
+      <div class="main-content main-content--two">
+        <div class="page-content page-content--intro">
+          {% block pageContentIntro %}
+            {% if intro.logo %}
+              {% set image = intro.logo %}
+              {% include "@atoms/09-media/image.twig" %}
```
gleroux02: I don't have a solution, but my preference would be to not print this image twice for performance reasons.

styleguide/source/_patterns/04-templates/01-content-types/services.twig
```diff
@@ -0,0 +1,80 @@
...
+            {% endif %}
+            {% set teaserText = intro.teaserText %}
+            {% include "@organisms/by-author/teaser-text.twig" %}
+            {% if intro.keyActions %}
+              {% set keyActions = intro.keyActions %}
+              {% include "@organisms/by-author/key-actions.twig" %}
+            {% endif %}
+          {% endblock %}
+        </div>
+        <aside class="sidebar sidebar--intro">
```
gleroux02: All of the elements inside of the aside are optional, let's make the aside markup optional too.

styleguide/source/_patterns/04-templates/01-content-types/services.twig
gleroux02: I don't think that this file should have the header or footer in it. As the system that implements it won't have that data at this level.

gleroux02 (changes requested):

styleguide/source/_patterns/04-templates/01-content-types/services.twig
gender404: There shouldn't be any performance issues if it is the exact same image (same path).  The Browser will only do one request to fetch the image and then display it in multiple locations.

gender404:
@gleroux02 - I think I fixed all of the issues you found.  I wasn't sure how to check if a block was empty so I added a flag to toggle the view between full and two column.

I am wondering if I need the blocks at all due to the structured nature of the content type.  thoughts?

gleroux02:
@legostud a way that you can check if a block exists would be like:
```
% set _block = block('sidebar') %}
{% if _block is not empty %}
{{ _block|raw }}
{% endif %}
```
Now do we need to have them? I need to think a bit longer on that.

gender404:
If I use that code, how to I add dummy content to the content type in Pattern Lab?  

I'm pretty sure sidebar would be considered empty and the content type would render full width instead of two columns.

gleroux02 (approved):

PR #376, opened 2017-03-21
DP-2471: GUIDE PAGE: Add a section with all available options

nstriedinger:
Please read: https://jira.state.ma.us/browse/DP-2471, build, and verify.


gender404:
@gleroux02 - should this be a new page Guide-Test or something like that so we can preserve the existing page as an example of a more real life use case?

gleroux02:
@legostud I can agree to that. I'm not sure what to call the new page. Something like "Guide - Section Example"?

gender404:
that's fine by me

PR #380, opened 2017-03-28
DP-2117 topic page update

jesconstantine:
This PR makes several updates to the Topics Page.

Ticket:
https://jira.state.ma.us/browse/DP-2117

Invision: https://projects.invisionapp.com/d/main#/console/10205887/217933611/preview

### Updated Patterns:

**Topic Page pattern updates:**
- Renamed the page pattern from L1 to TOPIC
- Added a link list to the bottom of the page for Related...
- Removed Common Actions component

**Page Banner pattern updates:**
- Added a new banner layout called "columns"
  - Background image spans right 2/3 columns
  - Description text spans left 1/3 columns
  - Banner title behaves similar to other page types

**Sections 3 Up pattern updates (card group):**
- Removed the optional compHeading
  - Seems like this heading could still be written semantically but only hidden visually.  Consideration for future work with headings: add a hidden property

**Section Links pattern updates (card):**
- Added a new optional variable linkType (defaults to decorative) to determine if the links should be decorative vs callouts
- Added the ability to render links as callouts
- Added a new optional variable seeAll (defaults to "") to determine if there should be a more link at the bottom of the card


jesconstantine:
Seems like we might want an option to make a compHeading visually hidden but semantically present.  Removing the heading all-together from the sections cards on this page seems like an accessibility concern to me.  Thoughts @ygannett ? @legostud ?

**This is what the design update includes:**
image



**This is what it used to be:**
image

gender404:
It's already missing on the landing page:
http://dev.massgov.velir.com/?p=pages-L0-visiting-and-exploring

That title is coded as optional

styleguide/source/_patterns/03-organisms/by-template/page-banner.twig
```diff
@@ -34,11 +36,30 @@
...
+        .ma__page-banner__image {
+          background-image: url('{{ pageBanner.bgNarrow }}');
+        }
+
+        @media (min-width: 801px) {
+          .ma__page-banner__image {
+            background-image: url('{{ pageBanner.bgWide }}');
+          }
+        }
+      </style>
```
gender404: I think I would try using an if else statement instead of having two if statements to control which element to apply the background images to.  This will keep them closer to each other and make it easier to understand what's going on.

{% if pageBanner.size == "columns" %}
<style>
  .ma__page-banner__image {
    background-image: url('{{ pageBanner.bgNarrow }}');
  }

  @media (min-width: 801px) {
    .ma__page-banner__image {
      background-image: url('{{ pageBanner.bgWide }}');
    }
  }
</style>
{% else %}
<style>
  .ma__page-banner {
    background-image: url('{{ pageBanner.bgNarrow }}');
  }

  @media (min-width: 801px) {
    .ma__page-banner {
      background-image: url('{{ pageBanner.bgWide }}');
   }
}
</style>
{% endif %}

jesconstantine: Thanks, addressed in https://github.com/massgov/mayflower/pull/380/commits/0eee6119a81165e96543b3dd36704b64f0bf6ab1

styleguide/source/_patterns/02-molecules/section-links.twig
```diff
@@ -6,6 +6,7 @@
 %}
 {% set titleClass = "ma__section-links__title js-accordion-link" %}
 {% set titleId = "section_titles_" ~ sectionLinks.id %}
+{% set linkType = sectionLinks.type ?: "decorative" %}
```
jesconstantine, line 9: sectionLinks can now render decorative links (this is what happens now, everywhere) or callout links (introduced on the cards on topic pages for services)

wondering if it makes more sense to check for sectionLinks.links (for decorative links) or sectionLinks.calloutLinks (for callouts) instead of passing a flag and keeping the data object the same (sectionLinks.links) -- thoughts @liberatr @jasonstanbery ?

styleguide/source/_patterns/02-molecules/callout-link.twig
```diff
@@ -1,4 +1,5 @@
-<a class="ma__callout-link" href="{{ calloutLink.href }}">
+{% set modifier = modifier ? " ma__callout-link--" ~ modifier : "" %}
+<a class="ma__callout-link{{ modifier }}" href="{{ calloutLink.href }}">
```
jesconstantine, line 2: When they appear in section links cards, callouts span the full width of their parent.  So I'm passing a modifier "full" in section-links.twig below prior to including this callout-link.twig

styleguide/source/assets/scss/03-organisms/_page-banner.scss
```diff
@@ -252,4 +298,39 @@ $page-banner-icon-width: $column + $gutter;
...
+    position: absolute;
+      right: 0;
+      top: 0;
+    width: 100%;
+
+    @media ($bp-medium-min) {
+      width: 62.12903%; // @include span(8 of 12) was too big
+      height: 460px;
+      margin-right: 1.6129%;
+      max-width: 820px;
```
jesconstantine, line 332: These large display image dimensions (820px by 460px) seem close enough to what we end up using for mobile in the page banner (800px by 400px) -- so I think this pattern only needs one image for the page banner.  Does that sound right?  We'd just adjust the display and center the image for mobile -- see line 321 above.

If that's the case, I'll update the <style> block and json data object accordingly.

jasonstanbery (approved):
This looks good and consistent with what we've been doing.

styleguide/source/_patterns/05-pages/TOPIC-state-parks-and-recreation.json
```diff
@@ -1,165 +1,133 @@
...
-      },
-      {
-        "href" : "#",
-        "text" : "More"
-      }]
-    }]
-  }
-}
+{
+  "urlPath": "/?p=pages-ACTION-get-a-state-park-pass",
```
gender404, line 2: The Action page was removed to be Topic instead

styleguide/source/_patterns/02-molecules/callout-link.twig
gender404: Couple issues here.  "modifier" is not name spaced so it could bleed into other components.  If you're going this approach you should have used a twig merge to added the this data to the calloutLink object.

That said, you don't need this modifier at all.  Since the calloutLink is contained with a Section Link, you can just add css to the section-link.scss file that targets ma__callout-link items and makes them {width: 100%:}

.ma__section-link {
 ...

  .ma__callout-link {
    width: 100%:
  }
}

styleguide/source/_patterns/02-molecules/section-links.twig
gender404: I would probably just skip this step and just use the value directly in the if statement below.

styleguide/source/_patterns/02-molecules/section-links.twig
```diff
@@ -32,12 +33,27 @@
         {% include "@atoms/decorative-link.twig" %}
       </div>
       <ul class="ma__section-links__items">
-        {% for decorativeLink in sectionLinks.links %}
-        <li class="ma__section-links__item">
-          {% include "@atoms/decorative-link.twig" %}
-        </li>
-        {% endfor %}
+        {% if linkType == "callout" %}
```
gender404, line 36: {% if sectionLinks.type == "callout" %}

styleguide/source/_patterns/02-molecules/section-links.twig
```diff
@@ -32,12 +33,27 @@
...
       </div>
       <ul class="ma__section-links__items">
-        {% for decorativeLink in sectionLinks.links %}
-        <li class="ma__section-links__item">
-          {% include "@atoms/decorative-link.twig" %}
-        </li>
-        {% endfor %}
+        {% if linkType == "callout" %}
+          {% for calloutLink in sectionLinks.links %}
+            {% set modifier = "full" %}
```
gender404, line 38: Not an ideal approach see comment above

styleguide/source/_patterns/02-molecules/section-links~callout-links.json
```diff
@@ -0,0 +1,27 @@
...
+    {
+      "href" : "#",
+      "text" : "Fishing"
+    },
+    {
+      "href" : "#",
+      "text" : "Hunting"
+    }]
+  }
+}
```
gender404, line 27: Line Endings - Greg hates these

styleguide/source/_patterns/03-organisms/by-template/page-banner.twig
```diff
@@ -34,11 +51,16 @@
           <span> {{ pageBanner.titleSubText }}</span>
         {% endif %}
       </h1>
-      {% if pageBanner.description %}
+      {% if pageBanner.description and ( pageBanner.size and not pageBanner.size == "columns" ) %}
```
gender404, line 54: Is check for existence required in twig ( pageBanner.size and not pageBanner.size == "columns" )?

I've been just doing if (pageBanner.size == "columns") in other places, but after seeing your code, I'm not sure which is better.

styleguide/source/assets/scss/02-molecules/_section-links.scss
```diff
@@ -39,6 +43,17 @@
...
+    display: none;
+
+    @media ($bp-x-small-min) {
+      align-self: flex-end;
+      border-top: 1px solid;
+      display: block;
+      padding: 0 20px 20px 0;
+      width: 100%;
+    }
+  }
```
gender404, line 56: I would reverse the media query.  Make these styles the default behavior and then do @media ($bp-x-small-max) { display: none; }  This well help with print styles that don't always respect min-width media queries and removes the need to do "display: block" as an override.

styleguide/source/assets/scss/02-molecules/_section-links.scss
```diff
@@ -131,6 +147,7 @@
 
   &__toggle-content {
     padding: 10px 0 30px;
+    flex: 1 1 auto;
```
gender404, line 150: you should break this out into three lines.  IE tends to has some weird issues with this.

styleguide/source/assets/scss/03-organisms/_page-banner.scss
```diff
@@ -9,7 +9,7 @@ $page-banner-icon-width: $column + $gutter;
   position: relative;
 
   @media ($bp-x-small-min) {
-    height: 366px;
+    min-height: 366px;
```
gender404, line 12: Not sure what this accomplishing.  you're overriding this with &--columns anyways.  This also breaks the default page banner size by making it 400px again.

styleguide/source/assets/scss/03-organisms/_page-banner.scss
```diff
@@ -90,10 +104,10 @@ $page-banner-icon-width: $column + $gutter;
   &--large &__content {
     height: 100%;
     max-width: 640px;
-    top: 0;
 
     @media ($bp-x-small-min) {
       margin-right: 105px;
+      top: 0;
```
gender404, line 110: Why was this moved?

styleguide/source/assets/scss/03-organisms/_page-banner.scss
```diff
@@ -111,6 +125,37 @@ $page-banner-icon-width: $column + $gutter;
     }
   }
 
+  &--columns &__content {
+    height: 400px;
+    max-width: 640px;
+    position: relative;
+    top: 200px;
+
+    &:after {
```
gender404, line 134: You don't need to solve this issue.  The gap is there on all of the other ones as well if the content isn't wide enough.

styleguide/source/assets/scss/03-organisms/_page-banner.scss
```diff
@@ -252,4 +298,39 @@ $page-banner-icon-width: $column + $gutter;
   &--overlay &__description {
     transform: skew(0);
   }
+
+  &--columns &__description {
+    font-size: 1.5rem;
+    margin-top: 0;
+    padding: 25px 0 20px;
+    transform: skew(0);
+    width: calc(100%);
```
gender404, line 307: just use width: 100%;  calc adds processing so only use it when you want to do math.

styleguide/source/assets/scss/03-organisms/_page-banner.scss
```diff
@@ -252,4 +298,39 @@ $page-banner-icon-width: $column + $gutter;
...
+  &--columns &__image {
+    background-position: 50% 50%;
+    height: 400px;
+    position: absolute;
+      right: 0;
+      top: 0;
+    width: 100%;
+
+    @media ($bp-medium-min) {
+      width: 62.12903%; // @include span(8 of 12) was too big
```
gender404, line 329: alphabetical order please.

styleguide/source/assets/scss/03-organisms/_page-banner.scss
gender404: that sounds like a good idea.  You can then simplify the code further, by just using an inline style tag on the __image div instead of the style block => style="background-image: url({{ ...}})".

styleguide/source/assets/scss/06-theme/03-organisms/_page-banner.scss
```diff
@@ -35,6 +34,18 @@ $page-banner-inner-width: 612px;
     background-image: none;
   }
 
+  &--columns &__content {
```
gender404, line 37: This is part of a larger issue that doesn't need to be addressed yet.

gender404 (commented):
Overall it looks pretty good, I'd like to see if we could find a way to fix some of these

styleguide/source/_patterns/02-molecules/callout-link.twig
jesconstantine: Thanks - I think I addressed this in the design QA branch, can you verify that this is better? https://github.com/massgov/mayflower/pull/385/commits/87e22d89a8cac16bba6ec55156dff93c29339d36

styleguide/source/assets/scss/03-organisms/_page-banner.scss
jesconstantine: @jasonstanbery and I went back and forth a few times here.  Ended up deciding that we will still give authors the option of adding a different image for mobile (could be completely different or using art direction on the same image)... with that in mind, does what we have here work?

styleguide/source/assets/scss/03-organisms/_page-banner.scss
jesconstantine: Left in by mistake, will remove.

styleguide/source/assets/scss/03-organisms/_page-banner.scss
gender404: What you have works fine.  

I just noticed the margin-right value.  I would remove that and correct the 'right' position to be 40px.

styleguide/source/assets/scss/03-organisms/_page-banner.scss
jesconstantine: Thanks - left in by mistake, will remove.

styleguide/source/assets/scss/03-organisms/_page-banner.scss
jesconstantine: Yep.  Will move.

styleguide/source/_patterns/02-molecules/callout-link.twig
gender404: Close, but I would do it the other way around.  I would add styles to the section-link.scss file to style the .ma__callout-link.  You added .ma__section-link to the callout-link.scss file.

There are pros and cons for either approach.  For me, I prefer to parents containing overrides for children.  

Pro: If the parent component is removed, then the overrides are also removed.  The child component is less likely to be retired.  This also allows us to use the child's Sass file on another site that may not have the same parent components. 

Con: It makes it a bit harder to update the child without breaking parents, because it isn't as obvious if a parent is overriding something.

styleguide/source/_patterns/03-organisms/by-template/page-banner.twig
jesconstantine: They way I'm interpreting the docs, I think it's required only if twig is configured with `strict_variables`: https://twig.sensiolabs.org/doc/2.x/templates.html#variables ... does that seem right?

See also: https://twig.sensiolabs.org/doc/2.x/api.html#environment-options

styleguide/source/assets/scss/03-organisms/_page-banner.scss
jesconstantine: The org page header was broken, the title text was off the screen, too far above.  I'll build from a prior tag and put a screenshot here.

styleguide/source/assets/scss/03-organisms/_page-banner.scss
jesconstantine: Makes sense.  Will do.  Addressed in https://github.com/massgov/mayflower/pull/393/commits/cc5557aa98a3dd4df4c3b871fe1c62f24e5035db

styleguide/source/_patterns/02-molecules/section-links.twig
jesconstantine: Address in https://github.com/massgov/mayflower/pull/393/commits/2ce39716a7c1d301d00922a3d639ac3f3b76444c

styleguide/source/_patterns/02-molecules/section-links.twig
jesconstantine: Addressed in https://github.com/massgov/mayflower/pull/393/commits/2ce39716a7c1d301d00922a3d639ac3f3b76444c

styleguide/source/_patterns/02-molecules/section-links.twig
jesconstantine: Addressed in https://github.com/massgov/mayflower/pull/385/commits/87e22d89a8cac16bba6ec55156dff93c29339d36

styleguide/source/_patterns/02-molecules/section-links~callout-links.json
jesconstantine: Addressed in https://github.com/massgov/mayflower/pull/393/commits/bcfbc30f283836772b22b81bd505ca4b9bb9e0b8

styleguide/source/_patterns/03-organisms/by-template/page-banner.twig
```diff
@@ -34,11 +51,16 @@
...
         </p>
       {% endif %}
     </div>
+    {% if pageBanner.description and ( pageBanner.size and pageBanner.size == "columns" ) %}
+      <p class="ma__page-banner__description">
+        {{ pageBanner.description }}
+      </p>
+    {% endif %}
   </div>
 </section>
```
jesconstantine, line 66: Line ending added in https://github.com/massgov/mayflower/pull/393/commits/bcfbc30f283836772b22b81bd505ca4b9bb9e0b8

styleguide/source/assets/scss/03-organisms/_page-banner.scss
jesconstantine: Removed in https://github.com/massgov/mayflower/pull/393/commits/bcfbc30f283836772b22b81bd505ca4b9bb9e0b8

styleguide/source/assets/scss/03-organisms/_page-banner.scss
jesconstantine: Removed in https://github.com/massgov/mayflower/pull/393/commits/bcfbc30f283836772b22b81bd505ca4b9bb9e0b8

styleguide/source/assets/scss/03-organisms/_page-banner.scss
jesconstantine: Addressed in https://github.com/massgov/mayflower/pull/393/commits/bcfbc30f283836772b22b81bd505ca4b9bb9e0b8

styleguide/source/assets/scss/02-molecules/_section-links.scss
jesconstantine: Addressed in https://github.com/massgov/mayflower/pull/393/commits/51a43cd3ec384d329f478ab0eff0af4b7fdd4652

styleguide/source/assets/scss/03-organisms/_page-banner.scss
jesconstantine: I thought I was matching how the org page banner displays, I didn't see the gap for that one.

styleguide/source/assets/scss/02-molecules/_section-links.scss
jesconstantine: Fixed in https://github.com/massgov/mayflower/pull/393/commits/0863b80825e81a1086e0a451d2c06043ce4108f9

styleguide/source/assets/scss/03-organisms/_page-banner.scss
gender404: The org page has a very long title, which is why the gap isn't visible.

jesconstantine: Removed in https://github.com/massgov/mayflower/pull/393/commits/0716ad7fc3a7ce0c593338abfaf92d7199b1756a

styleguide/source/_patterns/02-molecules/callout-link.twig
jesconstantine: Thanks for thinking out loud here.  I'm not sure which makes more sense to me, but I'm happy to stick with your convention.

Fixed in https://github.com/massgov/mayflower/pull/393/commits/0f18286e8d6cb322f8798f441440a443d3868417

styleguide/source/_patterns/05-pages/TOPIC-state-parks-and-recreation.json
jesconstantine: Fixed in https://github.com/massgov/mayflower/pull/393/commits/3382e00f4c4ac4bb5e1e13cde78ab54b145d03b4

PR #388, opened 2017-04-04
Change title and alt text of site logo for better pronunciation

GBirch:
This is the Mayflower part of https://jira.state.ma.us/browse/DP-2399

See https://github.com/massgov/mass/pull/608 for the Drupal part.

Apparently "Mass Gov" is pronounced "Mass Governor" by some screen readers.   This change fixes the problem for error and interstitial pages (and for any other page template in which it may subsequently be included).

Test: 
- [ ] On home page of patternlab confirm that the site logo in the header and footer have alt text that says "Mass.gov" in place of "Mass Gov" and that the title of the related link has the same replacement.
- [ ] Using a screen reader, confirm that when you tab to each logo, its name is pronounced "mass dot gov"


ygannett:
Confirmed that the changed item was pronounced as "mass dot gov" in Voiceover.

PR #389, opened 2017-04-04
Change aria presentation of + symbol on close button in utility nav

GBirch:
Makes the + sign inaudible to screen readers.

Completes https://jira.state.ma.us/browse/DP-2401

To test:
- [ ] navigate to /?p=organisms-utility-nav.  Using screen reader, enter the utility nav.  Tab around until you are on the close button, note that reader just says "Close", not "Close Plus"


ygannett:
The change confirmed.  Ready for merge.

PR #392, opened 2017-04-10
DP-2685 - How to fees

gender404:
Tabular data pattern created.  Displays as a table on wide screens and as a list on narrow screens.  breakpoint to switch happens at two different places based on 1-3 columns or 4+.

New Patterns:
1. Tabular Data Organism

Updated html and data:
1. Table Atom


gleroux02 (approved):
I'm not looking forward to the implementation code required to do this, but there is no good way around that. 👍

gender404:
yeah, flexible responsive tables was rough.

PR #403, opened 2017-04-17
DP-2734 - Regulation Page

gender404:
https://jira.state.ma.us/browse/DP-2734

New Regulations Detail page.

New Patterns:
1. Arrow Button
2. Arrow Nav
3. Back Button
4. Listing Table
5. Page Flipper


labbydev (commented):
@legostud it looks like the page is missing the breadcrumbs

gender404:
Nice catch.  I just pushed up a commit to add breadcrumbs

labbydev:
Thanks looks good now with the breadcrumbs!

PR #407, opened 2017-04-20
DP-3149 - Ordered Steps

gender404:
https://jira.state.ma.us/browse/DP-3149
Added new logic for when there is only one step to not show the step number


jesconstantine (approved):
Thanks @legostud ... funny because I was expecting to see a css change to remove the icon :before the ordered step if there was only 1.  Would there have been other alignment issues at that point?

Anyway, nice work. 👍

gender404:
simpler to get rid of the list html.  This removes the indentation and the number is one fell swoop.  Also it's probably more semantic, since lists should be greater than one item.

labbydev (approved):

PR #421, opened 2017-05-03
Update google-map.twig to add escape(js) filters

GBirch:
As entered in Drupal, address strings can contain characters that break the json passed to the google maps js.  This adds escaping to the address, and in an excess of caution, to the other textual output as well.

test:
[ ] No change to the marker output of http://localhost:3000/?p=pages-Service-Unemployment-Benefits
[ ] Add some crap to google-map.json infoWindow:address  (add some special characters, an apostrophe, what have you), confirm that marker output is not changed.


GBirch:
marker

PR #443, opened 2017-05-18
Dp 3190 tab order adjusted in header

mrossi113:
Adjusting the tab order in the header so that it focus on the utility nav before moving to the logo. 

JIRA https://jira.state.ma.us/browse/DP-3190

@legostud made the change you suggested in the JIRA ticket. I did test both desktop view and mobile view to see if it works and display the correctly.


styleguide/source/_patterns/03-organisms/by-template/header.twig
```diff
@@ -4,6 +4,9 @@
   <div class="ma__header__backto">
     <a href="http://www.mass.gov">Go to classic Mass.gov</a>
   </div>
+  <div class="ma__header__utility-nav">
+        {% include "@organisms/by-template/utility-nav.twig" %}
+      </div>
```
gender404: Please fix the indentation here.

styleguide/source/_patterns/03-organisms/by-template/header.twig
```diff
@@ -32,7 +35,7 @@
       <div class="ma__header__main-nav">
         {% include "@molecules/main-nav.twig" %}
       </div>
-      <div class="ma__header__utility-nav">
+      <div class="ma__header__utility-nav--mobile">
```
gender404: This is not how BEM works.  You're not allowed to have a modifier (--mobile) class without a base class.  This would be "ma__header__utility-nav ma__header__utility-nav--mobile" if you wanted to use a modifier.  

I would stay away from using the word "mobile" though since there are plenty of mobile devices that are 1200px wide.  "--narrow" would be more semantic so "ma__header__utility-nav ma__header__utility-nav--narrow"

You will also need to do a "--wide" modifier on the other utility nav "ma__header__utility-nav ma__header__utility-nav--wide".  This will make more sense after you read the CSS changes I've mentioned below.

styleguide/source/assets/scss/03-organisms/_header.scss
```diff
@@ -18,7 +18,25 @@ body {
...
+  @media (max-width: 840px) {
+  .ma__header__utility-nav {
+    display: none;
+  }
+
+  .ma__header__utility-nav--mobile, 
+  .ma__utility-nav__items{
+    display: block;
+    visibility: visible;
+  }
```
gender404: There are a few issues with this code:
1. This code should live within the existing "&__utility-nav" code to keep it all in the same place in case someone makes updates in the future.
2. The media query should be using the sass variable "$bp-header-toggle-max" to keep it in sync with all the other breakpoints (easier to adjust one variable in the future)
3. The indentation is not consistent which makes it harder to read
4. I would use two modifier classes "--narrow" and "--wide" to control the show hide functionality.

Below is what the CSS should look like for the utility-nav with these changes
```
  &__utility-nav {

    &--narrow {
      @media ($bp-header-toggle-min) {
        display: none;
      }
    }

    &--wide {
      position: absolute;
        top: 0;
      width: 100%;

      @media ($bp-header-toggle-max){
        display: none;
      }
    }
  }
```

gender404 (changes requested):

mrossi113:
@legostud made the changes you suggested from the review beforehand. Thank you for the feedback and I get what you mean about the BEM and CSS.

gender404 (approved):

PR #449, opened 2017-05-25
DP-3519 / DP-3520 location page

gender404:
https://jira.state.ma.us/browse/DP-3519
https://jira.state.ma.us/browse/DP-3520

I created a new content type for these pages based on the details content-type.  I wasn't 100% sure what the Drupal content type logic was for the location page so I left it using a large placeholders.  I can modify that though.

Changes:
1. New Location content type
2. Image Promos - Added optional title
3. Updated the General and Parks location pages to use the new template.

### DIFF
--- a/styleguide/source/_patterns/03-organisms/by-author/image-promos.json
+++ b/styleguide/source/_patterns/03-organisms/by-author/image-promos.json
@@ -1,5 +1,12 @@
 {
   "imagePromos": {
+    "compHeading": {
+      "title": "Image Promos",
+      "sub": false,
+      "color": "",
+      "id": "",
+      "centered": ""
+    },
     "items": [{
       "title": {
         "text": "Hiking & Walking",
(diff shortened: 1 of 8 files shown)


labbydev:
This LGTM, however I'm not sure what the changes to the Drupal content type are going to look like so I'm a little concerned about that.

labbydev (approved):

gleroux02:
Reviewing

gleroux02:
@legostud the current location page has the ability to display quick actions. With this change I don't see that option anymore if that is intended then we will need to alter the current location content type to remove those links.

Overall I don't have any issue with this PR. I think the current location page "should" still operate as it does currently as the underlying elements used to render currently have not changed. I'd suggest creating a new ticket that was to alter the current drupal node template / preprocess function used to render the location to use this new location template.

gender404:
Quick Actions are being removed from Mayflower in favor of using the Key Actions component in the header.

I can start a ticket for the Drupal upgrade.

gleroux02 (approved):

PR #459, opened 2017-06-02
DP-3191 Refactored 'Go to classic Mass.gov' link positioning

iansholtys:
https://jira.state.ma.us/browse/DP-3191

Adjusted styling on "Skip to main content" link to:
* Avoid overlap with the 'Go back to classic Mass.gov" link. Fixes the original issue outlined in DP-3191 by centering 'Skip to main content', which looks better visually as well.
* Make it visible on focus. Removes the need to leverage Drupal classes to do this.

### DIFF
--- a/styleguide/source/_patterns/03-organisms/by-template/header.twig
+++ b/styleguide/source/_patterns/03-organisms/by-template/header.twig
@@ -2,9 +2,7 @@
 <header class="ma__header" id="header">
   <a class="ma__header__skip-nav" href="#main-content">skip to main content</a>
   <div class="ma__header__backto">
-    <div class="ma__header__backto__container">
-      <a href="http://www.mass.gov">Go to classic Mass.gov</a>
-    </div>
+    <a href="http://www.mass.gov">Go to classic Mass.gov</a>
   </div>
   <div class="ma__header__utility-nav ma__header__utility-nav--wide">
     {% include "@organisms/by-template/utility-nav.twig" %}
--- a/styleguide/source/_patterns/03-organisms/by-template/header.twig
+++ b/styleguide/source/_patterns/03-organisms/by-template/header.twig
@@ -1,9 +1,9 @@
 <!-- Begin .header -->
 <header class="ma__header" id="header">
-  <a class="ma__header__skip-nav" href="#main-content">skip to main content</a>
   <div class="ma__header__backto">
     <a href="http://www.mass.gov">Go to classic Mass.gov</a>
   </div>
+  <a class="ma__header__skip-nav" href="#main-content">skip to main content</a>
   <div class="ma__header__utility-nav ma__header__utility-nav--wide">
     {% include "@organisms/by-template/utility-nav.twig" %}
   </div>
--- a/styleguide/source/_patterns/03-organisms/by-template/header.twig
+++ b/styleguide/source/_patterns/03-organisms/by-template/header.twig
@@ -2,7 +2,9 @@
 <header class="ma__header" id="header">
   <a class="ma__header__skip-nav" href="#main-content">skip to main content</a>
   <div class="ma__header__backto">
-    <a href="http://www.mass.gov">Go to classic Mass.gov</a>
+    <div class="ma__header__backto__container">
+      <a href="http://www.mass.gov">Go to classic Mass.gov</a>
+    </div>
   </div>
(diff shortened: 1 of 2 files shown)


styleguide/source/_patterns/03-organisms/by-template/header.twig
```diff
@@ -2,7 +2,9 @@
 <header class="ma__header" id="header">
   <a class="ma__header__skip-nav" href="#main-content">skip to main content</a>
   <div class="ma__header__backto">
-    <a href="http://www.mass.gov">Go to classic Mass.gov</a>
+    <div class="ma__header__backto__container">
```
gender404: One note:  BEM does not allow grandchildren.  ma__header__backto__container translate in BEM to namespace__parent__child__grandchild.  When you run into a situation where you need a grandchild that usually means you should create a new pattern.  In this case that would mean creating a new atom for the 'Go back to' button and then using an include to add it to the header inside the ma__header__backto div.

I don't actually believe you need to do this though since you can accomplish the same desired outcome by simply moving the skip link to be after the 'Go back to' button.

styleguide/source/assets/scss/03-organisms/_header.scss
```diff
@@ -31,6 +31,10 @@ body {
 
   &__skip-nav {
     @include ma-visually-hidden;
+    display: block;
+    margin: 0 auto;
+    width: 240px !important;
+    text-align: center;
```
gender404: Not sure why any of this worked for you with the @ma-visually-hidden mixin in there.  That mixin should have prevented the link from being visible in the browser.  Sounds like there may be a bug with this mixin not hiding content properly which could be a serious global issue.

gender404 (changes requested):
I see some overall issues with this PR and the Jira ticket.  

1. The "Skip Navigation" link is supposed to be visually hidden from users.  The Jira ticket and your code makes it sound like that is not happening consistently since you were able to style this link with the mixin still in the code so I feel there is a bug with the Sass ma-visually-hidden mixin.  That should have been the first thing to resolve.

2. If you do want to show the "Skip Navigation" link, I would suggest doing that only when it is in focus or more appropriately, hiding it when it's not in focus.  You can use CSS :not(:focus) pseudo selectors to accomplish this.

  &__skip-nav {
    display: block;
    margin: 0 auto;
    width: 240px;
    text-align: center;
   
    &:not(:focus) {
      @include ma-visually-hidden;
      width: 240px;
    }
  }

3. You then have the issue of the 'go back to' button being pushed down by the skip link.  A simple solution to that would be to swap the two links in the html.  I don't think it would be a terrible issue if people using keyboards hit the skip link after the back to link.

--

These changes will drastically reduce the amount of code changed and keep the skip link hidden (assuming you fix that bug).

styleguide/source/assets/scss/03-organisms/_header.scss
isaacchansky: Looks like perhaps you meant to remove the visually-hidden mixin? as is currently, this link is not visible to sighted users. If the desire is to show this link visually, I'd recommend the route @legostud mentioned in this PRs comments where you set the default to be the block styling and then apply the visually-hidden mixin with a `:not(:focus)` selector.

isaacchansky: @legostud  Just checked it out, the mixin is working properly

iansholtys:
Sorry for all the confusion on this one. The hidden link wasn't showing unexpectedly, rather, in Mayflower it wouldn't show up even with focus. The issue only surfaced when in Drupal, where additional classes were used to leverage Drupal's default classes for showing and hiding things on focus.

Based on your feedback and that from the rest of the team, it sounds like the appropriate fix is to make the 'Skip to main content' link work independently of Drupal. As such I've adjusted the PR as outlined above to make the link visible on focus.

As it turns out, centering the 'Skip to main content' link is sufficient to keep it from being overlapped by the "Back to..." link across all resolutions, so adjustments related to the "Back to" link have been removed.

iansholtys:
@legostud, I believe my updates address the issues you outlined. Could you give this another review?

gender404:
image

the "go back to" button is mis-aligned on desktop when the skip link becomes visible.  This is why I suggested swapping the order of those two items.

PR #477, opened 2017-06-19
DP-4603 Add prominent call to action for pilot feedback to all pilot pages

isaacchansky:
This PR creates a floating action molecule to allow a button to be added to the footer of every page and persist visually.

JIRA ticket:
https://jira.state.ma.us/browse/DP-4063

- Makes the back-to-top button optional.
- Adds a floating-action molecule which is also optional


styleguide/source/_patterns/02-molecules/floating-action.twig
```diff
@@ -0,0 +1,21 @@
...
+{% if floatingAction.theme %}
+  {% set buttonClass = buttonClass ~ "-" ~ floatingAction.theme %}
+{% endif %}
+
+<a
+    class="ma__button {{ buttonClass }} {{ buttonClassSize }} {{ buttonClassFloat }} {{ buttonClassAlignment }} {{ buttonClassColor }}"
+    href="{{ floatingAction.href }}"
+    title="{{ floatingAction.info }}">
+  {{ floatingAction.text }}
+</a>
```
gender404: This appears to be almost the same as the button.twig file.  I would suggest modifying button.twig file to change 'theme' to 'color' since that will make it more flexible.

I would then convert this to an include of button.twig wrapped in a div class="ma__floating-action".  Let the button styling be handled by ma__button-minor ma__button-minor--green and let the positioning be handled by the ma__floating-action

styleguide/source/assets/scss/02-molecules/_floating-action.scss
```diff
@@ -0,0 +1,29 @@
...
+
+  &--alignment-right {
+    right: 1.5625rem;
+  }
+  &--alignment-left {
+    left: 1.5625rem;
+  }
+
+  @media ($bp-small-max) {
+    position: fixed;
```
gender404: duplicate - not needed

styleguide/source/assets/scss/02-molecules/_floating-action.scss
```diff
@@ -0,0 +1,29 @@
+.ma__floating-action {
+  font-size: 1rem;
+  letter-spacing: .1em;
+  opacity: 1;
+  padding: .5em .7em;
+  position: fixed;
+    bottom: 1.3125rem;
+  transition: all .5s;
+  visibility: visible;
```
gender404: Not sure why opacity and visibility are here.  letter-spacing is usually only applied if the text is uppercase.

styleguide/source/_patterns/03-organisms/by-template/footer.twig
```diff
@@ -1,3 +1,6 @@
+{% if floatingAction %}
+  {% include "@molecules/floating-action.twig" %}
+{% endif %}
```
gender404, line 3: Can this exist at the same time as the back to top button?  You might need to add write some CSS .ma__floating-action ~ .ma__footer__back2top { top: ?? } to the footer.scss file

gender404 (changes requested):
Main concern is duplicating the ma__button code.  It would be better to update that pattern to allow it to be green.

The other concern is a lack of default styling.  The pattern should be able to be viewed without any --modifier classes. .class:not(.class--modifier) can be used if you want to do CSS as an if else statement.

gender404:
this feels like a tight deadline so I would suggest just adding 'color' as a new variable to button.twig and add it as an additional class {{ buttonClassColor }} that overrides the blue color for outline or solid.

isaacchansky:
@legostud made some changes to re-use existing atoms. I think ideally we want to probably not have both backtotop and a floated action in play at the same time, though theoretically you could float the action on the left and have backtotop on the right without an overlap issue. Seeing as they are both opt-in, I'm not sure it's worth writing more code to avoid that programatically.

styleguide/source/assets/scss/06-theme/01-atoms/_buttons.scss
```diff
@@ -20,3 +20,8 @@ button {
 .ma__button-minor-alt {
   @include ma-button("normal","outline",$c-theme-quaternary);
 }
+
+
+.ma__button-minor-green {
+  @include ma-button("normal","outline",$c-theme-secondary);
+}
```
gender404, line 27: somewhere we should note that this needs to be adjusted to be .ma__button-minor--green and .ma__button-minor--alt.  I'd say just change it, but I'm not sure if those classes are being used in Drupal.

styleguide/source/assets/scss/06-theme/02-molecules/_floating-action.scss
```diff
@@ -0,0 +1,14 @@
+.ma__floating-action {
+  z-index: $z-sticky-content;
```
gender404: I believe this should be in the other scss file

styleguide/source/assets/scss/06-theme/02-molecules/_floating-action.scss
```diff
@@ -0,0 +1,14 @@
...
+    background-color: $c-font-inverse;
+    border-color: $c-theme-secondary;
+    color: $c-theme-secondary;
+  }
+
+  &:hover {
+    background-color: $c-theme-secondary;
+    color: $c-font-inverse;
+  }
+}
```
gender404: I don't think this file is needed any more.

gender404 (changes requested):

styleguide/source/assets/scss/06-theme/01-atoms/_buttons.scss
gender404: I added a note to the button and button-link md file for this

gender404 (approved):

PR #485, opened 2017-06-22
DP-3854 - update label,placeholder,submit text for search

isaacchansky:
More accessible language in the search form

https://jira.state.ma.us/browse/DP-3854


styleguide/source/_patterns/03-organisms/by-template/header.twig
```diff
@@ -28,7 +28,7 @@
     <div class="ma__header__nav-container">
       {% if not hideHeaderSearch %}
         <div class="ma__header__nav-search">
-          {% set headerSearch = {'placeholder':'Search','id':'nav-search'} %}
+          {% set headerSearch = {'placeholder':'Search Mass.gov', 'label': 'Search terms','id':'nav-search'} %}
```
gender404: This should be modified now that we aren't using different copy for each of these.  I would suggest using a `|merge` to override the default id with the 'nav-search' value.

gender404 (changes requested):

styleguide/source/_patterns/03-organisms/by-template/header.twig
isaacchansky: 👍 thanks

gender404:
@isaacchansky - please update the title of this PR to include the Jira Ticket number.  DP-3854 - Update, ...

gender404 (approved):

PR #488, opened 2017-06-27
Switch Mayflower over to the Drupal Pattern Lab fork

sghoweri:
Includes multiple PL bug fixes merged in over the past couple of months, including the major fix for the native Pattern Lab Lineages functionality being broken when using Twig name-spaced paths for Drupal 8. 

Also updates several of the underlying Pattern Lab pieces (ex. Styleguidekit Assets Default) to the latest version, hence a few tiny UI changes like the styling of the tabs in the Pattern Lab "panel" UI getting updated.

**Side note:** I noticed a bunch of PHP Notices getting thrown but it seems like those were already happening before I folded in these updates. Besides that (and a ton of Composer dependencies being way out of date -- which I omitted from this PR) everything seems to be working great from what I can tell!

Before:
image

After:
image

CC @jesconstantine @bryanhirsch


gender404:
@sghoweri - thanks for the PR.  This will save me a bunch of time.

I'll review this shortly once we figure out how the contribution process works.

jesconstantine:
@sghoweri This is amazing!  Thank you so much!

styleguide/npm-shrinkwrap.json
```diff
@@ -1,6348 +1,5333 @@
...
-  "version": "2.0.0",
+  "version": "3.2.2",
+  "lockfileVersion": 1,
   "dependencies": {
     "abbrev": {
-      "version": "1.0.9",
-      "from": "abbrev@>=1.0.0 <2.0.0",
-      "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.0.9.tgz",
+      "version": "https://registry.npmjs.org/abbrev/-/abbrev-1.0.9.tgz",
+      "integrity": "sha1-kbR5JYinc4wl813W9jdSovh3YTU=",
```
gender404, line 8: I'm not sure why this updated itself, but it is a very different format which makes me a bit nervous.

gender404 (commented):
I should start this with a statement that I'm not a PHP developer so I may have missed some items.

Nothing I saw in here seemed like an issue except the npm-shrinkwrap.json file, but that seemed to work when I cloned the repo.  There were a couple of files with a mix of tabs and spaces, which is acceptable, but feels a bit sloppy.

The updated code does indeed resolve the issue with documentation not showing contains and included in information.

It would have been nice if it could have also figured out the includes and contains based on the below format as well, but I'm sure that is a much harder ask.
```
{% for element in richText.rteElements %}
    {% include element.path with element.data %}  
{% endfor %}   
```
An example of this would be the rich-text organism that uses a for loop to include elements.

gender404:
Well that was odd.  Apparently if you delete the target branch the PRs automatically close.

styleguide/config/config.yml
```diff
@@ -20,7 +20,8 @@ publicDir: public
 sourceDir: source
 defaultPattern: pages-readme2
 defaultShowPatternInfo: false
-lineageMatch: '{%([ ]+)?include( |\()[&quot;\&#039;]([A-Za-z0-9-_]+)[&quot;\&#039;](\))?(.*)%}'
+baseDir: /Users/sghoweri/mayflower/styleguide/
```
jesconstantine: @sghoweri is this `baseDir` config something custom that you use?  I ask because I don't see it in any of the default / example configs that I've looked at:
- [pattern-lab/patternlab-php-core](https://github.com/pattern-lab/patternlab-php-core/blob/master/config/config.yml.default)
- [drupal-pattern-lab/patternlab-php-core](https://github.com/drupal-pattern-lab/patternlab-php-core/blob/master/config/config.yml.default)
- [pattern-lab/patternlab-node](https://github.com/pattern-lab/patternlab-node/blob/master/patternlab-config.json)

Seems like we'd need to either remove this line, or make its path more relative to remove usernames and then make use of it - is it used similar to the jekyll baseDir?  Can you tell us how you use it for your projects? :)

sghoweri:
@jes

sghoweri:
@jesconstantine - derp. That might have gotten automatically added when I ran through a fresh Pattern Lab install... or I might have just goofed and copied something from my own PL config.

It feel like that might be an undocumented but still supported config option but in either case that’d definitely be something we’d strip out or make more generic — agreed.

Let me double-check when I get in this morning!

jesconstantine:
@sghoweri Thanks!  I was actually excited when I saw it.  So I'm going to poke around and see if / how other are using it!

sghoweri:
@jesconstantine check this out: https://github.com/pattern-lab/patternlab-php-core/search?utf8=✓&q=baseDir&type=

The `Config::getOption("baseDir")` instances mean that PL will look for that config option in the config yaml file if it exists (and a bunch do have defaults typically set)

jesconstantine:
@sghoweri gotcha thanks.  Based on the usage instances from your search it seems like its filesystem path for various core functionality, which makes sense given the value that was present from your prior commit.  

I was excited about the possibility of a config like jekyll's `baseurl` which allows you to serve from a given URL and can be used throughout templates as a config object property / variable.

styleguide/composer.lock
```diff
@@ -67,32 +66,33 @@
                 "tar",
                 "zip"
             ],
-            "time": "2016-02-15 22:46:40"
+            "time": "2016-02-15T22:46:40+00:00"
         },
         {
             "name": "doctrine/collections",
-            "version": "v1.3.0",
+            "version": "v1.4.0",
```
jesconstantine, line 73: @sghoweri - thanks for your patience as we review your PR.  This is a learning experience for us!

Since we see dependency package updates for doctrine (and also michelf, pear, phing, seld, symfony, and twig if we scroll through the `composer.lock` file below) and because we currently version our vendor directory, should we expect to see file system changes in `/styleguide/vendor` related to those packages in this PR?  

When I pull this branch and run `composer update pattern-lab/core` I see no updates to pattern-lab/core (makes sense since I believe those are committed already) but I do see updates to the packages mentioned above, which I see being updated here in the composer.lock file (and I don't see changes to composer.lock file which makes sense to me because the updates are already referenced here in the PR).  Am I thinking about this correctly? 

/cc: @weitzman @gleroux02 if you have the time to take a look and explain to me then we could save @sghoweri the time. :)

jesconstantine:
@sghoweri thanks again for your help here!  Your pattern inheritance will come in handy and we're excited to be working with the drupal pattern lab org projects!

gender404:
@sghoweri - I found an interesting bug that looks like it's related to this upgrade.  I have a variant `~` for our Contact List Organism that is only supposed to show one contact instead of the two shown in the base.  
http://mayflower.digital.mass.gov/?p=organisms-contact-list-single-item

On my local environment, I see two of contacts instead of one.  The first is from the variant json file and the second one is from the base json file.  I'm not sure how the data from the base is bleeding over into the variant.

sghoweri:
@legostud Hmmm. That is an strange bug indeed. 

I’m wondering if that could be related to the regex updates that pick up the twig @namespace syntax... or perhaps there’s a bug with the Pattern Lab data transform plugin (which lets you use the Drupal 8-friendly `{{ attributes }}` object... In any case, this is definitely something we should look into. Any chance you could add a bug to the Pattern Lab PHP Core repo on the Drupal Pattern Lab org? https://github.com/drupal-pattern-lab/patternlab-php-core/issues

In the meantime, as a possible workaround, could you try adding the `only` attribute to this include to see if that addresses the issue in this one spot (to force the data getting passed in to be encapsulated)?

```
 {% include "@molecules/contact-us.twig" only %}
```

https://github.com/massgov/mayflower/blob/dev/styleguide/source/_patterns/03-organisms/by-author/contact-list.twig#L12

Lemme know if that does the trick!

PR #495, opened 2017-06-29
DP-3195 Improve keyboard navigation of main menu

iansholtys:
This addresses a number of accessibility issues with the main menu:
* The buttons and links in the menu should all have `tabindex="-1"` except for the first one, which should be `0`
* Tab/Shift-Tab shouldn’t navigate through the menu items (that's what arrow keys are for, and it makes it take a very long time to skip past the menu)
* Menu bar and submenus should wrap around if you try to go past the end of a list
* Using the down or up arrows on a top-level menu item should expand that menu and move to the first or last items, respectively, rather than toggle the menu open and closed

With left/right and up/down arrows behaving so similarly, it was possible to consolidate and cut down on the code needed for those overall.

### DIFF
--- a/styleguide/source/_patterns/02-molecules/main-nav.twig
+++ b/styleguide/source/_patterns/02-molecules/main-nav.twig
@@ -5,15 +5,15 @@
       {% set menuId =  'menu' ~ loop.index %}
       <li class="ma__main-nav__item {{ nav.active ? 'is-active' : '' }} {{ nav.subNav ? 'has-subnav js-main-nav-toggle' : 'js-main-nav-top-link' }}">
         {% if nav.subNav %}
-          <button id="{{ buttonId }}" class="ma__main-nav__top-link" aria-haspopup="true" aria-expanded="false" aria-controls="{{ menuId }}">{{ nav.text }}</button>
+          <button id="{{ buttonId }}" class="ma__main-nav__top-link" aria-haspopup="true" aria-expanded="false" aria-controls="{{ menuId }}" tabindex="{{ loop.first? 0 : -1 }}">{{ nav.text }}</button>
           <div class="ma__main-nav__subitems js-main-nav-content is-closed">
             <ul id="{{ menuId }}" role="menu" aria-labelledby="{{ buttonId }}" class="ma__main-nav__container">
-              <li role="menuitem" class="ma__main-nav__subitem"><a href="{{ nav.href }}" class="ma__main-nav__link">{{ nav.text }}</a></li>
+              <li role="menuitem" class="ma__main-nav__subitem"><a href="{{ nav.href }}" class="ma__main-nav__link" tabindex="-1">{{ nav.text }}</a></li>
               {% for subNav in nav.subNav %}
-                <li role="menuitem" class="ma__main-nav__subitem"><a href="{{ subNav.href }}" class="ma__main-nav__link">{{ subNav.text }}</a></li>
+                <li role="menuitem" class="ma__main-nav__subitem"><a href="{{ subNav.href }}" class="ma__main-nav__link" tabindex="-1">{{ subNav.text }}</a></li>
               {% endfor %}
               <li role="menuitem" class="ma__main-nav__subitem">
-                <a href="{{ nav.href }}" class="ma__main-nav__link">
+                <a href="{{ nav.href }}" class="ma__main-nav__link" tabindex="-1">
                   {% include "@atoms/05-icons/svg-arrow-bent.twig" %}
                   <span>{{ nav.text }}</span>
                 </a>
@@ -21,7 +21,7 @@
             </ul>
           </div>
         {% else %}
-          <button id="{{ buttonId }}" class="ma__main-nav__top-link">{{ nav.text }}</button>
+          <button id="{{ buttonId }}" class="ma__main-nav__top-link" tabindex="{{ loop.first? 0 : -1 }}">{{ nav.text }}</button>
         {% endif %}
       </li>
     {% endfor %}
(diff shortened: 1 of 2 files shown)


styleguide/source/_patterns/02-molecules/main-nav.twig
```diff
@@ -5,23 +5,23 @@
...
+                <a href="{{ nav.href }}" class="ma__main-nav__link" tabindex="-1">
                   {% include "@atoms/05-icons/svg-arrow-bent.twig" %}
                   <span>{{ nav.text }}</span>
                 </a>
               </li>
             </ul>
           </div>
         {% else %}
-          <button id="{{ buttonId }}" class="ma__main-nav__top-link">{{ nav.text }}</button>
+          <button id="{{ buttonId }}" class="ma__main-nav__top-link" tabindex="{{ loop.first? 0 : -1 }}">{{ nav.text }}</button>
```
gender404, line 24: I don't understand the reasoning for the tabindex="-1".  If as you say the sub navigation should only be navigated with arrow keys then pressing tab should naturally proceed to the next main-nav item so you shouldn't need tab-index at all on the main-nav if you have tab-index="-1" on each of the sub-nav items

styleguide/source/assets/js/modules/mainNav.js
```diff
@@ -32,64 +32,45 @@ export default function (window,document,$,undefined) {
...
-          } else {
-            $dropdownLinks[focusIndexInDropdown-1].focus();
-            return;
+        // Up arrow was used
+        if(e.keyCode === 38) {
+          if(focusIndexInDropdown <= 1 ) {
+            focusIndexInDropdown = dropdownLinksLength;
           }
+          $dropdownLinks[focusIndexInDropdown-1].focus();
         } else {
```
legostud: needs a comment saying // Down arrow was used.  Otherwise, I naturally assumed any keyCode other than 38

styleguide/source/assets/js/modules/mainNav.js
```diff
@@ -102,44 +83,29 @@ export default function (window,document,$,undefined) {
...
-        //  - Open next pull menu and select first item
-        hide($openContent);
-        $topLevelLink.attr('aria-expanded','false');
-        let index = $topLevelLinks.index($topLevelLink)+1;
-        if($topLevelLinks[index]) {
-          $topLevelLinks[index].focus();
-        }
+        let leftArrow = e.keyCode === 37,
+            index = $topLevelLinks.index($topLevelLink),
+            linkCount = $topLevelLinks.length;
```
gender404: I prefer to declare my variables at the top of the section to make them easier to find.

styleguide/source/assets/js/modules/mainNav.js
```diff
@@ -102,44 +83,29 @@ export default function (window,document,$,undefined) {
...
+        index = ((index % linkCount) + linkCount) % linkCount;
+        $topLevelLinks[index].focus();
         return;
       }
 
-      // key code 9 is the tab key
-      if(open || (typeof(e.keycode) !== "undefined" && e.keycode !== 9)) {
+      if(open || (typeof(e.keycode) !== "undefined")) {
         return;
       }
```
gender404: I think this is leftover code that doesn't do anything.  A blank return is usually done to block the code that follows from running.  There isn't any code following this.

gender404 (changes requested):
The use of the arrow keys is much better than the tabs.  I think we have gone to far though by preventing the user the ability to tab through the top level Menu bar items without having to use the arrows.  I would allow that to happen naturally by only putting the tab-index="-1" on the submenu items.

Also, this is an existing issue, but we should make the keyboard functionality work for mobile by removing the below code in the keydown function.  On my laptop, I only have to zoom to 200% before the hamburger menu shows up so there is a use case for a keyboard being used for this menu style as well.

```
        // only for desktop
        return;
      }
```

gender404:
Also, I've mentioned this before, but I still feel that this code needs a spring cleaning, because it has been modified too many times by too many different people.  This could be a future Jira ticket.  

Ideally a few of us would sit down and record the desired behaviors to determine the proper JS class hooks needed and the proper events to listen too and then re-implement the code contained in a more organized and cleaner manner.

styleguide/source/_patterns/02-molecules/main-nav.twig
iansholtys: For this part I was mirroring the [W3C's Menubar example](https://www.w3.org/TR/wai-aria-practices/examples/menubar/menubar-1/menubar-1.html).

gender404: I can't argue with a spec. I still feel it's a poor UX to not be able to tab to each item in the main nav, but I'll let it pass.

gender404:
I found a couple of issues that will need to be resolved.  

1. If you open the last menubar item, once you down arrow to the last item, the page starts scrolling instead of looping back to the top. Steps: focus on the first menubar item, press the right arrow four times followed by the up arrow that should then prevent the arrow keys from working in Chrome on Win10.

2. We should either address keyboard navigation on narrow screens or disable it.  Currently the left and right arrow move up and down and the up and down arrows slide the menu left and right.  Please note that the narrow screen can be seen by zooming the browser in so this could be the view on a desktop (ie: has a keyboard) someone with poor vision could be using.

iansholtys:
@legostud, I made a number of fixes and changes here:
* Removed legacy behavior for the last item in the last submenu which was causing it to get stuck open and not wrap to the top of the menu
* Adjusted breakpoint to 840px to match the current CSS value
* Perform keycode-to-action mapping at the start of the function so the rest of the code is easier to understand
* Consolidate similar uses of preventDefault()
* Adjust functionality so when the mobile menu is in use, keypresses have the following behaviors:
** Up/down navigates through the main nav, or if a subnav is open, through that. Focus wraps at the ends of the menus.
** Right arrow expands a submenu for the current menu item.
** Left arrow (or escape) exits out of the the current submenu.

I believe this should take care of the issues you identified.

PR #496, opened 2017-06-30
DP-3911 form page UAT

gender404:
https://jira.state.ma.us/browse/DP-3911

Updates to input-text.twig
1. new maxlength field to restrict num characters entered
2. new pattern field to restrict the type of characters entered
3. If type number, leave as type number
4. if it has a js-input-number class, also add the same styling as type=number

Tooltip - removed hover and checked status for showing and hiding
Labels - increase font-weight from 500 to 700
Radio/checkbox inputs - adjusted alignment

### DIFF
--- a/styleguide/source/_patterns/01-atoms/03-forms/input-text.json
+++ b/styleguide/source/_patterns/01-atoms/03-forms/input-text.json
@@ -7,6 +7,8 @@
     "name": "text-input",
     "type": "text",
     "width": "",
+    "maxlength": "",
+    "pattern": "",
     "placeholder": "type something",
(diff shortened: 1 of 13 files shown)


gender404:
I feel this work is border line going too far into the realm of form validation which I don't feel belongs in Mayflower.  I tried to keep it fairly basic to show the intent, but stopped short of figuring out why the keycode restrictions don't work on my Android device.

isaacchansky:
I think that's the right move. Once we have more real-world examples we can revisit the client-side validation story.

labbydev (approved):

PR #499, opened 2017-06-30
DP-3228 Fixed use of aria-expanded in accordions

iansholtys:
When toggling accordions, `aria-expanded` wasn’t informing screen readers about the change. `aria-expanded` needs to apply to the button/link, not the element it’s toggling.

For accordion buttons, `aria-controls` was added to clarify the connection between the button and what it controls. If the accordion element doesn’t have an ID, one is generated and applied to allow this to work.

To get this change to apply to the Emergency Alert organism, I had to standardize the button markup so `.js-accordion-link` applies to the button, not the wrapper. To get this class to the button without disrupting other places which might use the Button Alert molecule, the class is added by passing it as a `class` property.

### DIFF
--- a/styleguide/source/_patterns/02-molecules/section-links.twig
+++ b/styleguide/source/_patterns/02-molecules/section-links.twig
@@ -8,11 +8,17 @@
         {% include "@atoms/05-icons/cat-icon.twig" %}
       </div>
     {% endif %}
-    {% set headingLevel = sectionLinks.subTitle ? 'h3' : 'h2' %}
-    <{{ headingLevel }} class="ma__section-links__title" id="{{ titleId }}">
-    {% set decorativeLink = sectionLinks.title|merge({ 'class' : 'js-accordion-link'}) %}
-      {% include "@atoms/decorative-link.twig" %}
-    </{{ headingLevel }}>
+    {% if sectionLinks.subTitle %}
+      <h3 class="ma__section-links__title js-accordion-link" id="{{ titleId }}">
+        {% set decorativeLink = sectionLinks.title %}
+        {% include "@atoms/decorative-link.twig" %}
+      </h3>
+    {% else %}
+      <h2 class="ma__section-links__title js-accordion-link" id="{{ titleId }}">
+        {% set decorativeLink = sectionLinks.title %}
+        {% include "@atoms/decorative-link.twig" %}
+      </h2>
+    {% endif %}
     <nav class="ma__section-links__toggle-content js-accordion-content" aria-labelledby="{{ titleId }}">
       {% if sectionLinks.description %}
         <p class="ma__section-links__description">{{sectionLinks.description}}</p>
--- a/styleguide/source/_patterns/02-molecules/section-links.twig
+++ b/styleguide/source/_patterns/02-molecules/section-links.twig
@@ -9,7 +9,7 @@
       </div>
     {% endif %}
     {% set headingLevel = sectionLinks.subTitle ? 'h3' : 'h2' %}
-    <{{ headingLevel }} class="ma__section-links__title" id="{{ titleId }}">
+      <{{ headingLevel }} class="ma__section-links__title" id="{{ titleId }}">
     {% set decorativeLink = sectionLinks.title|merge({ 'class' : 'js-accordion-link'}) %}
       {% include "@atoms/decorative-link.twig" %}
     </{{ headingLevel }}>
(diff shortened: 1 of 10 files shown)


styleguide/source/assets/js/modules/accordions.js
```diff
@@ -2,14 +2,16 @@ import checkActive from "../helpers/cssControlCode.js";
...
+  $('.js-accordion').each(function(index){
     let $el = $(this),
+        id = $el.attr('id') || 'accordion' + (index + 1),
         $link = $el.find('.js-accordion-link'),
         $content = $el.find('.js-accordion-content'),
         active = checkActive($el),
         open = $el.hasClass('is-open');
 
-    $el.attr('aria-expanded',open);
+    $el.attr('id', id);
```
gender404: I think this is supposed to be applied to the $content element.  The content that is being expanded and collapsed.
https://www.w3.org/WAI/GL/wiki/Using_the_WAI-ARIA_aria-expanded_state_to_mark_expandable_and_collapsible_regions

styleguide/source/_patterns/03-organisms/by-template/emergency-alerts.twig
```diff
@@ -4,8 +4,8 @@
       <div class="ma__emergency-alerts__container">
         {% set emergencyHeader = emergencyAlerts.emergencyHeader %}
         {% include "@molecules/emergency-header.twig" %}
-        <div class="ma__emergency-alerts__header-interface js-accordion-link">
-          {% set buttonAlert = emergencyAlerts.buttonAlert %}
+        <div class="ma__emergency-alerts__header-interface">
+          {% set buttonAlert = emergencyAlerts.buttonAlert|merge({ 'class': 'js-accordion-link' }) %}
```
gender404: I like this approach.  

Also, since this variable is used twice, I should have placed it outside of the `section` tag.  could you please place it between the `if emergencyAlerts.alerts` and the `section` tag.

gender404 (changes requested):
This looks good with just some minor tweaks.

Did you search all of the patterns for .js-accordion-link to make sure the class is also on the button for those patterns or to make sure these changes haven't broken another pattern?

styleguide/source/assets/js/modules/accordions.js
```diff
@@ -2,14 +2,16 @@ import checkActive from "../helpers/cssControlCode.js";
 
 export default function (window,document,$,undefined) {
 
-  $('.js-accordion').each(function(){
+  $('.js-accordion').each(function(index){
     let $el = $(this),
+        id = $el.attr('id') || 'accordion' + (index + 1),
```
gender404: if the id is supposed to be on the content then this would have to change as well to point to `$content.attr('id')`

iansholtys:
@legostud, I’ve reviewed all patterns where where `js-accordion-link` is used and haven’t found anything broken as a result of these changes.

I did encounter an instance where the fix didn’t apply, in the section-links molecule. This was a result of `js-accordion-link` being applied to a wrapper rather than directly to the button/link. As that link is generated with the decorative-link atom, I modified that pattern to add support for additional classes, similar to what was done for the button-alert molecule. While I was in there, I also did some minor code consolidation in the section-links molecule.

As you noticed, the button was supposed to reference the content that’s being toggled, not the overall widget, using `aria-controls`. This has been corrected as well.

styleguide/source/_patterns/02-molecules/section-links.twig
```diff
@@ -8,18 +8,12 @@
...
-    {% else %}
-      <h2 class="ma__section-links__title js-accordion-link" id="{{ titleId }}">
-        {% set decorativeLink = sectionLinks.title %}
-        {% include "@atoms/decorative-link.twig" %}
-      </h2>
-    {% endif %}
-    <nav class="ma__section-links__toggle-content js-accordion-body" aria-labelledby="{{ titleId }}">
+    {% set headingLevel = sectionLinks.subTitle ? 'h3' : 'h2' %}
+    <{{ headingLevel }} class="ma__section-links__title" id="{{ titleId }}">
+    {% set decorativeLink = sectionLinks.title|merge({ 'class' : 'js-accordion-link'}) %}
```
gender404: The problem here is that you've moved what is clickable.  The user use to be able to click the H3 to open and close the accordion, but now the user can only do so by clicking on the text, which is a smaller target area and doesn't include the plus sign.

I would suggest reverting this and creating a new ticket to address this issue, which will probably involve new html and css for the title.

gender404: A suggestion would be to have two links that the CSS toggles between.  One is the H3 with decorative link and the other is a button for the accordion based on the Contact List accordion style heading.

iansholtys:
I've confirmed the issue with the clickable area being smaller and have reverted that commit. Will post more notes on the ticket.

PR #504, opened 2017-07-06
DP-4080 - Versioning Documentation

gender404:
https://jira.state.ma.us/browse/DP-4080

Describing how Versioning works within Mayflower.


docs/versioning.md
```diff
@@ -0,0 +1,43 @@
...
+A minor version is done when new features are added to a pattern or a new pattern has been introduced into Mayflower.  These fixes should always be done in a backwards compatible manner to prevent them from causing issues with older date obects (.json).  It should be safe to pull in a minor versions, but you will need to make sure that your markup (twig/html) matches Mayflower's.
+
+If you wish to take advantage of the new features or patterns you may have to update some code (twig/html and or data object) on your site.
+
+Example:
+
+1. Creating a new atom, molecule, or organism.
+2. Expanding a pattern's twig file to include new markup (twig/html)
+3. Modifying markup in a pattern's twig file.
+4. Adding new variables to a pattern's data object (.json or .twig).
```
jesconstantine: Are there ways to do this that are cause for both Minor and Major versions?

For example, would conditional / non-required new variables be considered Minor and required new variables be considered Major?

gender404: Well, adding new variables to the data object will always be fine.  It's the implementation of those variables in the markup that could cause a breaking change.

gender404:
I pushed a second draft

docs/versioning.md
```diff
@@ -0,0 +1,59 @@
...
+
+The Mayflower Style Library is using Semantic Versioning to help maintain dependencies with consumers of the patterns contained within.  This method will help convey what will be required to upgrade a site to a newer version of Maylfower.
+
+You can read more about Semantic Versioning [here](http://semver.org/)
+
+## Terms used in this document:
+1. Markup - The twig or html code for a pattern
+2. CSS - The generated index and base-theme css file generated from the SASS files.
+3. JS - The generated index.js JavaScript file
+4. Data Oject - The json data used for a pattern
```
isaacchansky: *Object

docs/versioning.md
```diff
@@ -0,0 +1,59 @@
...
+
+
+## The Basics
+Versions on Mayflower are denoted in by three values 1.0.0 that represent Major, Minor, and Patch changes respectively.  Major changes can contain Minor changes which can contain Patch changes.  When Major and Minor releases are made the numbers to the right will always be reset to zero.
+
+### Patch versions (0.0.1) 
+A patch version is created when an issue with a pattern in Mayflower has been resolved.  The changes made will be done in a backward compatible manner that doesn't require changes to markup or data objects.
+
+#### How to upgrade
+Patch versions will require an update the CSS or JS files.  In order to get all of the fixes, you may have to update your markup to match Mayflower's, but a patch will always work with older markup from the same minor release.  The data-object will never need to be updated.
```
isaacchansky: >"you may have to update your markup to match Mayflower's, but a patch will always work with older markup from the same minor release" 

This seems a little contradictory at first glance - is it equivalent to say "a patch will never visually break code using existing markup from the same minor release"?

docs/versioning.md
```diff
@@ -0,0 +1,59 @@
+# Semantic Versioning
+
+The Mayflower Style Library is using Semantic Versioning to help maintain dependencies with consumers of the patterns contained within.  This method will help convey what will be required to upgrade a site to a newer version of Maylfower.
```
isaacchansky: *Mayflower

docs/versioning.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Terms used in this document:
+1. Markup - The twig or html code for a pattern
+2. CSS - The generated index and base-theme css file generated from the SASS files.
+3. JS - The generated index.js JavaScript file
+4. Data Oject - The json data used for a pattern
+
+
+## The Basics
+Versions on Mayflower are denoted in by three values 1.0.0 that represent Major, Minor, and Patch changes respectively.  Major changes can contain Minor changes which can contain Patch changes.  When Major and Minor releases are made the numbers to the right will always be reset to zero.
```
isaacchansky: *Versions of Mayflower are denoted by...

isaacchansky (changes requested):

docs/versioning.md
jesconstantine: I think that's also worth an example with version numbers.  We want this document to be as clear and understandable as possible, in order to mitigate any version / update implementation risk.

docs/versioning.md
```diff
@@ -0,0 +1,66 @@
...
+### Terms used in this document
+* Markup - The twig or html code for a pattern
+* CSS - The Style Sheet index-generated.css and base-theme-generated.css files
+* JS - The JavaScript index-generated.js and vendor-generated.js files
+* Data Object - The JSON data used for a pattern
+
+
+## Our Contract with you
+We will make sure that the Data Object for all patterns, except pages, works consistently between Major versions.  
+*- Pages are there as examples to show different ways to combining patterns.*
```
jeffmaher: Suggestion:

Mayflower patch and minor versions will not introduce changes to markup, CSS, JS, and Data Object APIs/contract. Patches releases existing functionality that is broken. Minor releases add new functionality, without modifying the contract of the existing features. Consumers can safely upgrade to newer patch and minor versions without having to a thorough regression test of their code.

Major versions can change anything. It's expected that consumers upgrading to a major version need to retest areas of their code that utilize Mayflower.

jesconstantine: RE: minor versions, I also like what @isaacchansky was saying about the open / close principle: patterns are open for extension but closed for modification.  So a markup/css/js change that extends current functionality but doesn't modify it would be okay.

For example if we wanted to introduce the practice of using `data-attr` vs `.js-class` for js hooks in markup, then the minor version way of doing it would be to release code wth both `data-attr` and `.js-class` in the markup while the major way of doing it would be to replace `.js-class` with `data-attr`

@plusjeff how does that resonate with your comment above that minor versions will not introduce changes to markup (and even css / js in other examples)

@legostud how does this relate to how you've been thinking about and drafting minor versions?

jeffmaher: @jesconstantine That sounds great!

gender404: As a little insight, which I'll add to the contract, is that my assumption is that the consumer will always being pulling in the latest JS and CSS bundles for their current Major release.  Possibly doing this from a CDN that provides a version for each Major release.  This way the only 'changes' the consumer will have to do is updated markup, extend the data object, or upgrade everything to a new Major release.

gender404:
new version uploaded for review.  I need to find better examples of Patch releases now, because the one I used isn't valid anymore.

docs/versioning.md
```diff
@@ -0,0 +1,64 @@
...
+- *Patch:* consumer doesn't need to re-write any Markup, things just get better, nothing new is available, nothing breaks just by updating
+- *Minor:* consumer doesn't need to re-write any Markup, has access to new features available, maybe an option to do something a better way, nothing breaks just by updating
+- *Major:* consumer has to walk through a 'migration' step to update since this will break things
+
+## The Basics
+Versions of Mayflower are denoted by three values 1.0.0 that represent Major, Minor, and Patch changes respectively.  Major changes can contain Minor changes which can contain Patch changes.  When Major and Minor versions are made the numbers to the right will always be reset to zero.
+
+### Key Terms
+* Markup - The twig or html code for a pattern
+* Artifacts - The SVG icons, fonts, CSS, and JavaScript files
```
jesconstantine: What do you think about using Assets or even Static Assets here?

docs/versioning.md
```diff
@@ -0,0 +1,64 @@
...
+
+[5.3.1](https://github.com/massgov/mayflower/pull/497) - Was an incorrect Patch version.  We found a mistake in the Data Object where one of the new patterns introduced used `url` in the Data Object and `href` in the Markup.  This seemed like a small fix to correct the Data Object, but this should have been a Major version instead since we required the new variable in the Data Object.  This should have been done in a backward compatible manner that checks for the old variable and uses that if it exists.  
+
+
+### Minor versions (0.1.0)
+A Minor version is created when new features are added to an existing Pattern, a new Pattern is created, or an issue has been resolved that required Markup changes. New features will always be written as an optional extention of the existing pattern without requiring any modifications to the existing Markup or Data Object.
```
jesconstantine: I think 

> an issue has been resolved that required Markup changes

needs to be called out in the "Our Contract with the consumer" section, currently:

> *Minor:* consumer doesn't need to re-write any Markup, has access to new features available, maybe an option to do something a better way, nothing breaks just by updating

Suggestion:

> *Minor:* consumer only needs to re-write Markup in order to fix a broken pattern, has access to new features available, maybe an option to do something a better way, nothing breaks just by updating

Thoughts?

docs/versioning.md
```diff
@@ -0,0 +1,64 @@
...
+* Additional Markup or Data Object for a new feature added to an existing Pattern (ie: an optional title added to the Link List organism).
+* Existing Markup enhanced to work with a new optional feature (ie: new 'centered' variable for the title).
+* Broken Markup, CSS, or JS changed to resolve an issue (ie: typo found in a class name, twig variable, or CSS selector)
+
+#### Example version
+[5.4.0](https://github.com/massgov/mayflower/pull/507) - We introduced several new patterns and resolved a few issues with existing patterns.  The new patterns were introduced to facilitate a new Executive Order and Policy Advisory pages.  We also resolved several issues we found in the previous 5.3 version.
+
+
+### Major versions (1.0.0)
+A Major version is created when the new code requires changes to the Markup or Data Object to prevent breaking existing styling or functionality.
```
jesconstantine: Can you clarify

> ... to prevent breaking existing styling or functionality

Does it mean if you don't update your markup or data object then your site will break?

docs/versioning.md
gender404: I think I want to say "doesn't need to re-write any Markup unless specifically stated in the release notes, ...".   The expectation is that for the majority of Minor releases they shouldn't have to re-write any Markup.

Thoughts?

docs/versioning.md
gender404: yes, it means your site will break.  Maybe I change it to state that and then add another sentence describing what "break" could mean.

docs/versioning.md
jesconstantine: That works for me!

docs/versioning.md
gender404: Assets works for me.

docs/versioning.md
gender404: I changed this to "A Major version is created when the new code will break existing code unless updates are made to the Markup or Data Object."

jesconstantine (approved):
Nice work here @legostud !!

gender404:
@plusjeff - any comments?

@isaacchansky - did I fix the changes requested and does this work for you?

jeffmaher:
@legostud : This looks great! Thanks for writing up!

PR #506, opened 2017-07-11
Trigger Browser Sync reload after Pattern Lab build

EvanLovely:
I **have not** tested this, but I bet this would work. I saw @legostud mention [here](https://github.com/drupal-pattern-lab/roadmap/issues/3#issuecomment-312676451) that this was an issue, so I took a look and the [browser sync docs on the CLI command reload](https://github.com/drupal-pattern-lab/roadmap/issues/3#issuecomment-312676451) might be what you're looking for. I [handle it a bit different](https://github.com/theme-tools/theme-tools/blob/master/packages/plugin-pattern-lab-php/pattern-lab.js#L23) in my tools. 

Cheers! Happy Pattern Labbing!

### DIFF
--- a/styleguide/tools/gulp/tasks/patternlab.js
+++ b/styleguide/tools/gulp/tasks/patternlab.js
@@ -9,8 +9,9 @@ var gulp         = require('gulp'),
 
 module.exports = function patternLabTask(config, env){
 
+    var bsPort = config.local.browserSyncPort || 3000;
     gulp.task('patternlab', function() {
-        run('php ' + config.patternLabRoot + '/core/console --generate --patternsonly').exec();
+        run('php ' + config.patternLabRoot + '/core/console --generate --patternsonly && browser-sync reload --port ' + bsPort).exec();
     });
 
     // gulp.task('patternlab', function(next) {


jesconstantine:
@EvanLovely I'll let @legostud review since it's in response to his comment, but I wanted to pop in and say thank you so much!  Also, we're excited to be using, following + interacting with the Drupal Pattern Lab projects :)

EvanLovely:
Welcome! Glad to hear we've got some fans out there and the tools are helpful!

gender404:
Thanks @EvanLovely.  This appears to work great and doesn't have the same issue I ran into before.

I had originally tried the below code, but when I switched branches, Gulp basically nuked itself due to the number of 'pattern' updates it saw and would fire dozens of browserSync reloads causing an overload.

~~~
run('php ' + config.patternLabRoot + '/core/console --generate --patternsonly').exec(function() { 
            browserSync.reload(); 
        })
~~~

EvanLovely:
@legostud Glad it works! If that was the issue, or becomes one in the future, consider setting the [`reloadDelay` option](https://browsersync.io/docs/options#option-reloadDelay) to something like a half second. I think I have a default of 200 milliseconds anyway - it seems silly that the default is 0.

Anyway; glad it worked! Happy Pattern Labbing! 🎉

PR #509, opened 2017-07-12
DP-3234 Added visually-hidden h1 prefix content

iansholtys:
Although categories above `<h1>` tags are readily apparent to sighted users, they don't translate as well to assistive technologies. To address this, the category is now hidden from screen readers, and is added (but visually-hidden) at the start of the `<h1>`.

Because illustrated headers display the category differently and don't leverage the `pageHeader.category` property, a `pageHeader.prefix` property is added to provide a heading prefix in the absence of a category.

### DIFF
--- a/styleguide/source/_patterns/03-organisms/by-template/page-header.twig
+++ b/styleguide/source/_patterns/03-organisms/by-template/page-header.twig
@@ -1,6 +1,4 @@
-{% if pageHeader.widgets %}
-  {% set widgetsClass = "ma__page-header--has-widgets" %}
-{% endif %}
+{% set widgetsClass = pageHeader.widgets ? "ma__page-header--has-widgets" : "" %}
 {% set prefix = pageHeader.category|default(pageHeader.prefix) %}
 
 <section class="ma__page-header {{ widgetsClass }}">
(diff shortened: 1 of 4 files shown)


styleguide/source/_patterns/03-organisms/by-template/page-header.twig
```diff
@@ -1,6 +1,7 @@
 {% if pageHeader.widgets %}
   {% set widgetsClass = "ma__page-header--has-widgets" %}
 {% endif %}
```
gender404: Could you please update this code to be `{% set widgetsClass = pageHeader.widgets ? "ma__page-header--has-widgets" : "" %}`.  I know this is outside of the scope of the ticket, but it's good practice to clean up poorly written code as it's found to prevent future bugs (variable is optional, but global so previous pattern could set it).

styleguide/source/_patterns/03-organisms/by-template/page-header.twig
```diff
@@ -25,13 +26,16 @@
...
-        <div class="ma__page-header__category">
+        <div class="ma__page-header__category" aria-hidden="true">
           {{ pageHeader.category }}
         </div>
       {% endif %}
       <h1 
       class="ma__page-header__title"
-      {% if schemaPageType %} property="name"{% endif %}>{{pageHeader.title}}</h1>
+      {% if schemaPageType %} property="name"{% endif %}>
+        {% if prefix %}<span class="visually-hidden">{{ prefix }} </span>{% endif %}
```
gender404: you might want to use `&nbsp;` after {{ prefix }} just to make sure no one accidentally deletes the white space.

gender404 (changes requested):
Looks pretty good.  I put in a couple suggested changes.

iansholtys:
@legostud Good thoughts, I've made the requested changes and pushed those up.

PR #510, opened 2017-07-13
DP-4211 add iframe template

ygannett:
JIRA: [DP-4211 \[dev\] Build iframe functionality](https://jira.state.ma.us/browse/DP-4211)

- Add _iframe.twig_ to **/styleguide/source/_patterns/01-atoms/09-media/** along with _iframe.json_ and _iframe.md_.
- Add __iframe.scss_ to **/styleguide/assets/scss/01-atoms/**, add it to _index.scss_.

Note:

- iframe is rendered as a child of _pageContent_ in location detail and service detail pages.
- The template is created based on /styleguide/source/_patterns/01-atoms/09-media/video.twig.
- The scss file is created based on /styleguide/assets/scss/01-atoms/_video.scss.
- Its associated PR for Drupal: [DP-4211 [dev] Build iframe functionality #948](https://github.com/massgov/mass/pull/948)


gender404:
I started reviewing this until I realized that it was just a copy of the video.twig file with a couple of tweaks so most of my review notes were actually against code that I had written.  The code for the video pattern wasn't up to the current standards so I upgraded that component as part of this ticket and then updated the new iframe pattern to match those changes.

With the new iframe pattern, we should be able to phase out the video pattern in favor of this more generic pattern for iframes in the future.

Also, since the iframe pattern is a copy of the video pattern, I noticed that you hadn't copied the Theme scss file that applies the blue background.  I made that change as well assuming that these were supposed to match.  If this isn't supposed to have the background color, then we'll need to remove the padding: 13px from the container along with the background-color.

ygannett:
@legostud Thank you so much.  I'll get this ready for an external review.

PR #511, opened 2017-07-13
DP-3225 Aria role for Wait Time

mrossi113:
To update the "Wait Time" with the aria role `role="status" aria-live=“polite”` to let the screen reader know when a change to the wait time has occurred. 

JIRA ticket https://jira.state.ma.us/browse/DP-3225

Testing:
This can't be tested in Mayflower for we don't have wait time being updated.

I used the live site RMV page `https://pilot.mass.gov/locations/natick-rmv-service-center` and added the `role="status" aria-live="polite"` into the inspector tool to test. I used Chrome and VoiceOver

Below are the screenshots:
 screen shot 2017-07-13 at 3 22 48 pm

screen shot 2017-07-13 at 3 23 12 pm

### DIFF
--- a/styleguide/source/_patterns/02-molecules/wait-time.twig
+++ b/styleguide/source/_patterns/02-molecules/wait-time.twig
@@ -1,4 +1,4 @@
-<section class="ma__wait-time">
+<section class="ma__wait-time" role="status" aria-live=“polite”>
   <h2 class="ma__wait-time__title">
     {% include "@atoms/05-icons/svg-wait-time.twig" %}
     <span>Wait Time</span>


jesconstantine:
@mrossi113 thanks for this update!  We could update the actual wait time app so the component is more accessible for anyone who implements it - what do you think?  Your could PR would be opened on that repo: https://github.com/massgov/rmvwaittime

mrossi113:
@jesconstantine that sounds good! I will work on this today.

mrossi113:
Leaving this open until we have figure out if the `rmvwaittime` repo will work with these changes in Drupal instance.

mrossi113:
Closing this PR for not needed anymore. This work was done in the https://github.com/massgov/rmvwaittime/pull/5

PR #518, opened 2017-07-24
DP-3232 & DP-1369 Key Actions and Key Agencies showing multiple times throughout page

mrossi113:
For accessibility reasons added an aria-labelledby to the key actions section and used the compHeading.id to identify each of the key actions section. This is also happening in the key agencies section on the sidebar. 

JIRA ticket: https://jira.state.ma.us/browse/DP-3232 & https://jira.state.ma.us/browse/DP-1369

The key actions are being used both on the Guide (harden-content type) and the Stacked Layout (prototype - current homepage) at the moment. Used the compHeading.id instead of compHeading.title not to duplicate the key action title.

### DIFF
--- a/styleguide/source/_patterns/01-atoms/04-headings/sidebar-heading.twig
+++ b/styleguide/source/_patterns/01-atoms/04-headings/sidebar-heading.twig
@@ -1,1 +1,1 @@
-<h3 class="ma__sidebar-heading" aria-labelledby="{{compHeading.id}}">{{sidebarHeading.title}}</h3>
+<h3 class="ma__sidebar-heading">{{sidebarHeading.title}}</h3>
--- a/styleguide/source/_patterns/01-atoms/04-headings/sidebar-heading.twig
+++ b/styleguide/source/_patterns/01-atoms/04-headings/sidebar-heading.twig
@@ -1,1 +1,1 @@
-<h3 class="ma__sidebar-heading">{{sidebarHeading.title}}</h3>
+<h3 class="ma__sidebar-heading" aria-labelledby="{{compHeading.id}}">{{sidebarHeading.title}}</h3>
--- a/styleguide/source/_patterns/03-organisms/by-author/key-actions.twig
+++ b/styleguide/source/_patterns/03-organisms/by-author/key-actions.twig
@@ -8,7 +8,7 @@
       {% set sidebarHeading = keyActions.sidebarHeading %}
       {% include "@atoms/04-headings/sidebar-heading.twig" %}
     {% endif %}
-    <div class="ma__key-actions__items" aria-labelledby="{{compHeading.id}}">
+    <div class="ma__key-actions__items" aria-labelledby="">
       {% for link in keyActions.links %}
         {% if link.image %}
           {% set illustratedLink = link %}
--- a/styleguide/source/_patterns/03-organisms/by-author/key-actions.twig
+++ b/styleguide/source/_patterns/03-organisms/by-author/key-actions.twig
@@ -1,14 +1,15 @@
+{# backward compatible for version 5.5 #}
+{% if keyActions.sidebarHeading %}
+  {% set keyActions = keyActions|merge({"compHeading": { "title": keyActions.sidebarHeading.title, "sidebar" : true }}) %}
+{% endif %}
+
 <section class="ma__key-actions">
   <div class="ma__key-actions__container">
     {% if keyActions.compHeading %}
       {% set compHeading = keyActions.compHeading %}
       {% include "@atoms/04-headings/comp-heading.twig" %}
     {% endif %}
-    {% if keyActions.sidebarHeading %}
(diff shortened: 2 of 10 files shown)


styleguide/source/_patterns/03-organisms/by-author/key-actions.twig
```diff
@@ -8,7 +8,7 @@
       {% set sidebarHeading = keyActions.sidebarHeading %}
       {% include "@atoms/04-headings/sidebar-heading.twig" %}
     {% endif %}
-    <div class="ma__key-actions__items">
+    <div class="ma__key-actions__items" aria-labelledby="{{compHeading.id}}">
```
gender404: if compHeading is set to null and a sidebarHeading is used instead this will fail.

styleguide/source/_patterns/01-atoms/04-headings/sidebar-heading.twig
```diff
@@ -1 +1 @@
-<h3 class="ma__sidebar-heading">{{sidebarHeading.title}}</h3>
\ No newline at end of file
+<h3 class="ma__sidebar-heading" aria-labelledby="{{compHeading.id}}">{{sidebarHeading.title}}</h3>
```
gender404: You should never ever reference variables from another namespace (compHeading vs sidebarHeading).  If I was to include a pattern with a Comp Heading containing an ID every Sidebar Heading would inherit that same value. 
 And IDs are not allowed to be duplicated on a page.

To solve this you should add an 'id' value to the sidebarHeading which defaults to null.  

CC: @jesconstantine

gender404 (changes requested):
This ticket leans towards some improvements that I was hoping to do.  Maybe this change could be incorporated as part of this fix.

What I'd like to do is retire the Sidebar Heading, by adding a 'sidebar' variable to the Comp Heading that when set to true would render the Comp Heading as a sidebar.  This variable would add a modifier class (ma__comp-heading--sidebar) to the Comp Heading.  The scss would be updated to the following:
~~~
.ma__comp-heading {

  &:not(&--sidebar) {
    @include ma-comp-heading;
  }

  &--sidebar,
  .sidebar & {
    @include ma-sidebar-heading;
  }

  &--centered {
    text-align: center;

    &:after {
      left: 50%;
      transform: translateX(-50%);
    }
  }

  .sidebar--colored & {
    @include ma-heading-colored($direction: left);
    border-bottom: none;

    &:after {
      display: block;
    }
  }
}
~~~

styleguide/source/assets/scss/01-atoms/_comp-heading.scss
```diff
@@ -10,14 +19,6 @@
     }
   }
 
-  .sidebar & {
-    @include ma-sidebar-heading;
-
-    &:after {
```
gender404: I think the &:after is still going to be needed, but can be moved to within the `&--sidebar, .sidebar & {` selector.

mrossi113: @legostud moved the `&after` within the `&--sidebar, .sidebar & {` which works fine in the http://localhost:3000/?p=atoms-comp-heading. But the http://localhost:3000/?p=atoms-comp-heading-centered `&after` is left align now in this twig. I tried a few things, but it won't position the bar correctly in the center. Not sure what to do with this?

gender404: There were two &:after { ... } blocks and you had the one for centering inside the .sidebar.  I moved that, but it still didn't work due to the :not() adding more specificity so I fixed that as well and it should be centering now.

gender404:
I found a few other issues in there when I was working on this.  

1. If the sidebar variable was set to true, two headings were being rendered.  
2. The new sidebar variable wasn't added to the compHeading.json file which is the source of truth for which variables are used.
3. The compHeading.md file wasn't updated with the new variable
4. A variant compHeading~as-sidebar-heading.json wasn't created.

I pushed a fix to correct those.

Do this code still meet your requirements?

styleguide/source/_patterns/03-organisms/by-author/key-actions.twig
```diff
@@ -8,7 +8,7 @@
       {% set sidebarHeading = keyActions.sidebarHeading %}
       {% include "@atoms/04-headings/sidebar-heading.twig" %}
     {% endif %}
-    <div class="ma__key-actions__items">
+    <div class="ma__key-actions__items" aria-labelledby="">
```
gender404: I'm pretty sure this should be populated.

gender404: I updated this to use keyActions.compHeading.id

mrossi113:
@legostud Yes, the code changes you made meets all of the requirements for both of the JIRA tickets. Thank you for making the changes and helping me with this ticket!

PR #519, opened 2017-07-25
Workaround to address unexpected Pattern Lab Psuedo-Pattern Data Inheritance

sghoweri:
- Reverting the current Pattern Lab psuedo-pattern rules to instead continue to use the original merge function vs array_replace_recursive as a temporary workaround to address unexpected data inheritance / merging behavior raised in https://github.com/drupal-pattern-lab/patternlab-php-core/issues/12

- Adding the composer patch library to automatically apply this array_merge patch if/when upstream PL updates should occur, until patch is no longer needed (ie. when we can figure out the best way to customize this sort of data merging behavior -- perhaps via https://github.com/drupal-pattern-lab/patternlab-php-core/issues/11) 

On a side note: I'd strongly encourage you guys to consider omitting the composer vendor folder from version control (for the same reasons why including the node_modules folder is considered to be a no no) -- that would help with reducing overall diff noise in addition to ensuring composer dependencies are always most up to date (which is partially why this "unexpected behavior" from Pattern Lab is an issue, given that these upstream changes occurred over a year ago).

CC @jesconstantine @legostud @EvanLovely @aleksip


jesconstantine:
Thanks again @sghoweri I'll review this shortly!

EvanLovely:
I would also encourage the `vendor/` folder being added to `.gitignore` and taken out of version control. Sometimes servers don't have the ability to build a site (i.e. run `composer install` or `npm install`) and we've handled that by having a second "deploy repo" that everything is copied into, committed, and then pushed.

jesconstantine (approved):
@sghoweri this works great for me!!

@legostud my functional tests were:
- [short link list data](http://localhost:3000/?p=organisms-link-list-short) 
- [split columns 2 data](/?p=organisms-split-columns-two)

...which were two problem-child patterns that I was aware of.

jesconstantine:
@EvanLovely re: not versioning vendor, the you describe is actually what we do for the [massgov/openmass](https://github.com/massgov/openmass) project (which is our deploy repo), we just hadn't caught Mayflower up to speed -- but this is a great kick in the pants for us to do it, so thanks!

gender404 (approved):

gender404 (changes requested):
@sghoweri - were going to merge this in to C4-Documentation and then cherry pick it into C4-docs.

PR #522, opened 2017-07-27
DP-3235 Added aria-label for file download links

iansholtys:
Similar to the change made in PR 487, but with the inclusion of the file size (if provided) in the aria-label per external reviewer feedback

### DIFF
--- a/styleguide/source/_patterns/02-molecules/download-link.twig
+++ b/styleguide/source/_patterns/02-molecules/download-link.twig
@@ -13,9 +13,9 @@
       {% include "@atoms/decorative-link.twig" %}
     {% else %}
       <a 
-        class="ma__download-link__file-link" 
-        aria-label="Open {{ downloadLink.format }} file{% if downloadLink.size %}, {{ downloadLink.size }},{% endif %} for {{ downloadLink.decorativeLink.text }}"
+        class="ma__download-link__file-link"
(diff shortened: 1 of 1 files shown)


styleguide/source/_patterns/02-molecules/download-link.twig
```diff
@@ -14,6 +14,7 @@
     {% else %}
       <a 
         class="ma__download-link__file-link" 
+        aria-label="Open {{ downloadLink.format }} file{% if downloadLink.size %}, {{ downloadLink.size }},{% endif %} for {{ downloadLink.decorativeLink.text }}"
```
gender404: Instead of using aria-label on an `a` tag, it would be better to just use a `title` attribute.  Titles will provide information to everyone not just screen readers.  Note: title only works on `a` tags for screen readers.

gender404 (commented):
minor change that I would prefer.

iansholtys:
In this instance I'm not sure there's a need to expose the information to other devices using `title` as it should all be readily available to sighted users. Also, when using OS X's VoiceOver utility, the title is read after the link's normal text, which isn't very helpful.

Because we're just prefacing the normal link text (the name of the file) with some additional information (the action, and file type/size), the simpler approach here is to just use the `visually-hidden` class on a span inside the link. This should behave consistently across all screen readers.

PR #526, opened 2017-07-31
DP-4103 - Add `align-self` style to section title to fix wrapping.

edwardchan:
This PR adds an update to add the `align-self` style to the section titles, which fixes the issue happening on IE11.

https://jira.state.ma.us/browse/DP-4103

After the update, the titles should wrap inside the content divs:
dp-4103


gender404:
technically, we only needed to add width: 100% to fix this in IE11, but this solution also works.

edwardchan:
Thanks @legostud !

PR #529, opened 2017-08-01
DP-4178 topic page

gender404:
This PR provides documentation for every pattern used on the Topic Page that hasn't already be documented:
http://mayflower.digital.mass.gov/?p=pages-topic
http://mayflower.digital.mass.gov/?p=pages-topic-your-government

NOTE: This PR includes the changes found in https://github.com/massgov/mayflower/pull/521, https://github.com/massgov/mayflower/pull/523, https://github.com/massgov/mayflower/pull/524, and https://github.com/massgov/mayflower/pull/527 so please merge those PRs first.

Jira:
https://jira.state.ma.us/browse/DP-4178
https://jira.state.ma.us/browse/DP-4177

Documentation was added to these patterns (with some twig changes):
Topic Page
* Single Column Template
* Section Three Up
  * Section Links
    * Cat Icon

I also removed the unnecessary documentation for SVG icons

PR #527, opened 2017-08-01
DP-4178 Location page

gender404:
This PR provides documentation for every pattern used on the Location Page that hasn't already be documented:
http://mayflower.digital.mass.gov/?p=pages-location-general-content
http://mayflower.digital.mass.gov/?p=pages-location-park-content

NOTE: This PR includes the changes found in https://github.com/massgov/mayflower/pull/521, https://github.com/massgov/mayflower/pull/523, and https://github.com/massgov/mayflower/pull/524 so please merge those PRs first.

Jira:
https://jira.state.ma.us/browse/DP-4178
https://jira.state.ma.us/browse/DP-4177

Documentation was added to these patterns (with some twig changes):
Location Page
* Location Template
  * Header Alert
  * Page Header - new variant for this page type
    * Location Icons
  * Location Banner
* Google Map
* Image Promos
  * Image Promo
* Event Listing
  * Event Teaser
  * Link (fixing some twig logic)

Other Fixes
* Button - fixed link in documentation
* Button - fixed empty modifier being added
* Link - fixed logic to switch from string to boolean

PR #524, opened 2017-07-28
DP-4178 / DP-4177 Guide Page

gender404:
This PR provides documentation for every pattern used on the Guide Page that hasn't already be documented:
http://mayflower.digital.mass.gov/?p=pages-guide

NOTE: This PR includes the changes found in https://github.com/massgov/mayflower/pull/521 and https://github.com/massgov/mayflower/pull/523 so please merge those PRs first.

Jira:
https://jira.state.ma.us/browse/DP-4178
https://jira.state.ma.us/browse/DP-4177

Documentation was added to these patterns (with some twig changes):

Guide Page
* Stacked Row Template
  * Emergency Alerts
    * Emergency Header
    * Button Alert
    * Emergency Alert
      * Link (changed data variables to use info instead of label and made chevron a boolean)
* Illustrated Header
* Jump Links (added missing title attribute)
* Stacked Row Section
* Suggested Pages
* Callout Stats
* Callout Time
* Callout Alert

PR #523, opened 2017-07-28
DP-4178 / DP-4177 How To page

gender404:
This PR provides documentation for every pattern used on the How To Page that hasn't already be documented:
http://mayflower.digital.mass.gov/?p=pages-howto

NOTE: This PR includes the changes found in https://github.com/massgov/mayflower/pull/521 so please merge that PR first.

Jira:
https://jira.state.ma.us/browse/DP-4178

Documentation was added to these patterns (with some twig changes):

HowTo Page
* Details Template
    * Page Header
        * Header Tags
        * Social Links
        * Publish State
        * Divider
    * Colored Heading
    * Sticky Nav
* Tabular Data
  * Table
* Steps Unordered
  * Action Step
* Steps Ordered
* Contact List
  * Contact Us
    * Contact Group (removed structured data elements)
    * Column Heading
* Link List

* Rich Text - fixing wording on documentation

PR #533, opened 2017-08-02
DP-4178 event page

gender404:
This PR provides documentation for every pattern used on the Event Page that hasn't already be documented:
http://mayflower.digital.mass.gov/?p=pages-event
http://mayflower.digital.mass.gov/?p=pages-event-recurring

NOTE: This PR includes the changes found in https://github.com/massgov/mayflower/pull/521, https://github.com/massgov/mayflower/pull/523, https://github.com/massgov/mayflower/pull/524, https://github.com/massgov/mayflower/pull/527, and https://github.com/massgov/mayflower/pull/529 so please merge those PRs first.

Jira:
https://jira.state.ma.us/browse/DP-4178
https://jira.state.ma.us/browse/DP-4177

Documentation was added to these patterns (with some twig changes):
Event / Event Recurring
* Event Template
  * Page Header - event variant
  * Page Overview
  * Event Listing - post content variant
  * Labelled List


styleguide/source/_patterns/03-organisms/by-author/page-overview.md
```diff
@@ -0,0 +1,29 @@
+### Description
+This Pattern is used to explain the reason for the content of a page.
```
jesconstantine, line 2: I'm not sure I understand this pattern - Is it different than rich text with title?  Also, it seems like it could be used more often than a page "overview", right?

styleguide/source/_patterns/02-molecules/labelled-list.md
```diff
@@ -1,25 +1,18 @@
----
-title: Labelled List
```
jesconstantine, line 2: Unrelated to this PR, but for my own learning, why is this not an `@atoms/lists/` pattern?  I see there is a parent `<div>` in the template, is that the only reason?

styleguide/source/_patterns/03-organisms/by-author/event-listing.md
```diff
@@ -12,6 +12,7 @@ A list of events where each item displays the Name, Location, Time, Date and a b
 
 ### Variant options
 * Can be used in the [Right Rail](./?p=organisms-event-listing-as-sidebar)
+* Renders as two columns with a background in [Post Content](./?p=organisms-event-listing-as-post-content)
```
jesconstantine: 💯 nice.

jesconstantine (approved):
Thanks @legostud !! I'm curious about a couple patterns (noted below) which we can discuss and if necessary plan future work-- I think this is good to go!

styleguide/source/_patterns/02-molecules/labelled-list.md
gender404: It's mainly here, because I'm still trying to properly define atom, molecule, and organism.  Even Brad Frost contradicts himself based on his  demo Pattern Lab.  Atoms are supposed to be a single html element, but a Table (`table`, `tr`, `td`, `tbody` ... ) is still an Atom?  It feels like it should be in with Atoms, but based on strict documentation it's actually a Molecule.

styleguide/source/_patterns/03-organisms/by-author/event-listing.md
gender404: hmm, I think I need to update this to use a template for the variant to show how it actually appears

styleguide/source/_patterns/03-organisms/by-author/page-overview.md
gender404: It contains more than just the Rich Text (labelled list, and Image).  I think it use to have a video as well.  All of these need to flow properly when the image is floated to the right.

gender404:
I think I've corrected everything or left comments

jesconstantine:
Good to go @legostud !! 👍

PR #534, opened 2017-08-02
DP-4198: Changes a tag to div to fix google translate.

smurrayatwork:
**Description:** 

This pull request changes the anchor tags used to contain the markup within callout-link.twg to be a div instead. Without this change, google translate ends up duplicating the for every callout link on the homepage, breaking the Featured Actions section.

Making this change is also more consistent with other links that are rendered on the page that use divs.

**Jira:** 
https://jira.state.ma.us/browse/DP-4198

**To Test:**
- [ ] Go to the homepage and scroll down to the Featured Actions section. Verify that the Featured Actions section renders properly.
- [ ] Scroll back up and use the dropdown labeled "Select Language" to choose any other language than English. Scroll back down to the Featured Actions section and verify that the Featured Actions section is still rendered properly.

**Post Deploy Steps:**
N/A

**Relevant Screenshots/GIFs:**
N/A

**Additional Notes:**
N/A
---

### DIFF
--- a/styleguide/source/_patterns/02-molecules/callout-link.twig
+++ b/styleguide/source/_patterns/02-molecules/callout-link.twig
@@ -1,7 +1,8 @@
-<a class="ma__callout-link" href="{{ calloutLink.href }}">
+<div class="ma__callout-link">
   <span class="ma__callout-link__container">
     <span class="ma__callout-link__text">
-      {{ calloutLink.text }}{% if calloutLink.info %}<span class="ma__callout-link__info">{{ calloutLink.info }}</span>{% endif %}&nbsp;{% include "@atoms/05-icons/svg-arrow.twig" %}
+      <a href="{{ calloutLink.href }}">{{ calloutLink.text }}</a>
+      {% if calloutLink.info %}<span class="ma__callout-link__info">{{ calloutLink.info }}</span>{% endif %}&nbsp;{% include "@atoms/05-icons/svg-arrow.twig" %}
     </span>
   </span>
-</a>
+</div>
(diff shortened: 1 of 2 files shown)


gender404 (changes requested):
This is broken.

1. The SVG arrow icon is no longer part of the link
2. The Icon wraps to a new line (orphans) for longer text (I'm assuming this based on the code)
3. You've pretty much reverted the changes made for DP-1521

Also since the Callout Link is pretty much the same structure as a Decorative Link, do we have a problem with Decorative Links on the site as well.

image

isaacchansky:
As for point 3 above while it might change the markup back to what it was like, the issue before was JS hi-jacking the anchor tag, so if this acts just as a normal anchor I think we're good.

smurrayatwork:
@legostud I've put the svg arrow back into the anchor tag. Does this address your concerns, here? If not, let me know. Thanks!

gender404:
@smurrayatwork - that doesn't appear to have fixed the issue with the arrows (see below).  Also now the entire box isn't clickable although the mouse pointer still says it is.

It does appear that you've found a bug with the arrows wrapping to a new line too early.  This doesn't appear to be happening in Prod, but I can see it in Dev.  Apparently the new hidden info span tag between the arrow and the text was adding extra white space causing the arrow to wrap.

I think a better approach to this is to wrap the a tag in a div.  

image

gender404:
See new PR to include those changes => https://github.com/massgov/mayflower/pull/535

gender404 (approved):

PR #535, opened 2017-08-03
DP-4198 - Callout Link - wrapping it in a div for google translate

gender404:
Wrapping the Callout Link in a DIV as a potentially better solution than making just the text clickable.


smurrayatwork:
@legostud This change works for google translate and doesn't break the layout - approved. Thanks!

smurrayatwork (approved):

PR #539, opened 2017-08-07
add headerTags to pageBanner

labbydev:
Include the headerTags as options to the pageBanner

New patterns:
* Content Eyebrow organism (http://localhost:3000/?p=organisms-content-eyebrow)

### DIFF
--- a/styleguide/source/_patterns/03-organisms/by-template/content-eyebrow.json
+++ b/styleguide/source/_patterns/03-organisms/by-template/content-eyebrow.json
@@ -1,0 +1,39 @@
+{
+  "contentEyebrow": {
+    "headerTags": {
+      "label": "More about:",
+      "taxonomyTerms": [{
+        "href": "#",
+        "text": "Term 1"
+      },{
+        "href": "#",
+        "text": "Term 2"
+      },{
+        "href": "#",
+        "text": "Term 3"
+      },{
+        "href": "#",
+        "text": "Term 4"
+      }]
+    },
+    "socialLinks": {
+      "label": "Share:",
+      "items": [{
+        "altText": "Follow us on Facebook",
+        "href": "#",
+        "icon": "@atoms/05-icons/svg-facebook.twig",
+        "linkType": "facebook"
+      },{
+        "altText": "Follow us on Twitter",
+        "href": "#",
+        "icon": "@atoms/05-icons/svg-twitter.twig",
+        "linkType": "twitter"
+      },{
+        "altText": "Follow us on LinkedIn",
+        "href": "#",
+        "icon": "@atoms/05-icons/svg-linkedin.twig",
+        "linkType": "linkedin"
+      }]
+    }
+  }
+}
--- a/styleguide/source/_patterns/03-organisms/by-template/content-eyebrow.json
+++ b/styleguide/source/_patterns/03-organisms/by-template/content-eyebrow.json
@@ -1,5 +1,6 @@
 {
   "contentEyebrow": {
+    "hideBorder": false,
     "headerTags": {
       "label": "More about:",
       "taxonomyTerms": [{
--- a/styleguide/source/_patterns/03-organisms/by-template/content-eyebrow.twig
+++ b/styleguide/source/_patterns/03-organisms/by-template/content-eyebrow.twig
@@ -1,5 +1,5 @@
 {% if contentEyebrow.headerTags.taxonomyTerms or contentEyebrow.socialLinks.items %}
-  <section class="ma__content-eyebrow{{contentEyebrow.hideBorder ? '--borderless' : ''}}">
+  <section class="ma__content-eyebrow {{contentEyebrow.hideBorder ? 'ma__content-eyebrow--borderless' : ''}}">
     <div class="ma__content-eyebrow__tags">
       {% if contentEyebrow.headerTags.taxonomyTerms %}
         {% set headerTags = contentEyebrow.headerTags %}
--- a/styleguide/source/_patterns/03-organisms/by-template/content-eyebrow.twig
+++ b/styleguide/source/_patterns/03-organisms/by-template/content-eyebrow.twig
@@ -1,0 +1,14 @@
+<section class="ma__content-eyebrow">
+  {% if contentEyebrow.headerTags.taxonomyTerms or contentEyebrow.socialLinks.items %}
+    <div class="ma__content-eyebrow__tags">
+      {% if contentEyebrow.headerTags.taxonomyTerms %}
+        {% set headerTags = contentEyebrow.headerTags %}
+        {% include "@molecules/header-tags.twig" %}
+      {% endif %}
+      {% if contentEyebrow.socialLinks.items %}
+        {% set socialLinks = contentEyebrow.socialLinks %}
+        {% include "@molecules/social-links.twig" %}
+      {% endif %}
+    </div>
+  {% endif %}
+</section>
--- a/styleguide/source/_patterns/03-organisms/by-template/content-eyebrow.twig
+++ b/styleguide/source/_patterns/03-organisms/by-template/content-eyebrow.twig
@@ -1,5 +1,5 @@
 {% if contentEyebrow.headerTags.taxonomyTerms or contentEyebrow.socialLinks.items %}
-  <section class="ma__content-eyebrow">
+  <section class="ma__content-eyebrow{{contentEyebrow.hideBorder ? '--borderless' : ''}}">
     <div class="ma__content-eyebrow__tags">
       {% if contentEyebrow.headerTags.taxonomyTerms %}
         {% set headerTags = contentEyebrow.headerTags %}
--- a/styleguide/source/_patterns/03-organisms/by-template/content-eyebrow.twig
+++ b/styleguide/source/_patterns/03-organisms/by-template/content-eyebrow.twig
@@ -1,5 +1,5 @@
-<section class="ma__content-eyebrow">
-  {% if contentEyebrow.headerTags.taxonomyTerms or contentEyebrow.socialLinks.items %}
+{% if contentEyebrow.headerTags.taxonomyTerms or contentEyebrow.socialLinks.items %}
+  <section class="ma__content-eyebrow">
     <div class="ma__content-eyebrow__tags">
       {% if contentEyebrow.headerTags.taxonomyTerms %}
         {% set headerTags = contentEyebrow.headerTags %}
@@ -10,5 +10,5 @@
         {% include "@molecules/social-links.twig" %}
       {% endif %}
     </div>
-  {% endif %}
-</section>
+  </section>
+{% endif %}
--- a/styleguide/source/assets/scss/03-organisms/_content-eyebrow.scss
+++ b/styleguide/source/assets/scss/03-organisms/_content-eyebrow.scss
@@ -1,0 +1,38 @@
+
+.ma__content-eyebrow {
+  @include ma-container;
+  @include clearfix;
+
+  &__tags {
+    clear: both;
+    display: flex;
+      align-items: flex-start;
+      flex-wrap: wrap;
+      justify-content: space-between;
+    overflow: hidden;
+    margin-bottom: -20px;
+    padding-top: 18px;
+
+    .ma__header-tags,
+    .ma__social-links {
+      border-bottom-width: 1px;
+      border-bottom-style: solid;
+      min-height: 48px;
+      position: relative;
+        top: 1px;
+    }
+
+    .ma__header-tags {
+      flex-grow: 1;
+      margin-bottom: 30px;
+    }
+
+    .ma__social-links {
+      flex-grow: 0;
+
+      @media ($bp-small-max) {
+        width: 100%;
+      }
+    }
+  }
+}
--- a/styleguide/source/assets/scss/03-organisms/_content-eyebrow.scss
+++ b/styleguide/source/assets/scss/03-organisms/_content-eyebrow.scss
@@ -35,4 +35,14 @@
       }
     }
   }
+
+  &--borderless {
+    @include ma-container;
+    @include clearfix;
+
+    .ma__header-tags,
+    .ma__social-links {
+      border: none;
+    }
+  }
 }
--- a/styleguide/source/assets/scss/03-organisms/_content-eyebrow.scss
+++ b/styleguide/source/assets/scss/03-organisms/_content-eyebrow.scss
@@ -36,9 +36,7 @@
     }
   }
 
-  &--borderless {
-    @include ma-container;
-    @include clearfix;
+  &--borderless &__tags {
 
     .ma__header-tags,
     .ma__social-links {
--- a/styleguide/source/assets/scss/index.scss
+++ b/styleguide/source/assets/scss/index.scss
@@ -109,6 +109,7 @@
 @import "03-organisms/change-log";
 @import "03-organisms/collage-header";
 @import "03-organisms/contact-list";
+@import "03-organisms/content-eyebrow";
 @import "03-organisms/emergency-alerts";
 @import "03-organisms/error-page";
 @import "03-organisms/event-listing";
--- a/styleguide/source/_patterns/04-templates/01-content-types/services.json
+++ b/styleguide/source/_patterns/04-templates/01-content-types/services.json
@@ -1,5 +1,6 @@
 {
   "contentEyebrow": {
+    "hideBorder": false,
     "headerTags": {
       "label": "More about:",
       "taxonomyTerms": [{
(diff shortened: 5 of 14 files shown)


styleguide/source/_patterns/03-organisms/by-template/page-banner.twig
```diff
@@ -2,6 +2,16 @@
 
 {% set colorClass = pageBanner.color ? "ma__page-banner--" ~ pageBanner.color : '' %}
 
+{% if pageBanner.headerTags.taxonomyTerms %}
+  <div class="ma__page-banner__tags">
```
gender404: This is an anti-pattern since you can't have a child `ma__page-banner__tags` element located outside of a parent element `ma__page-banner`.

gender404 (changes requested):
I think the overall request here is incorrect.  The ticket shouldn't be to added header tags to the Page Banner,  It should be to add them to the site Header instead since they can exist on any page regardless of which Page Header/Banner used.

I would suggest creating a new 'Page Navigation' Pattern (or some other name) by spinning off the the code used in the Page Header for these tags and Social Links.

Doing it this way will allow header tags on any page.  It will also allow us to place the header tags outside of the #main element making it possible for screen readers to skip these links as well when they click 'skip navigation'

styleguide/source/assets/scss/03-organisms/_page-banner.scss
```diff
@@ -323,4 +323,29 @@ $page-banner-icon-width: $column + $gutter;
       width: 62.12903%; // @include span(8 of 12) was too big
     }
   }
+
+  &__tags {
```
gender404: I think you forgot to also create the 'theme' version (colors, and font-weights) of this code.

styleguide/source/_patterns/03-organisms/by-template/page-banner.twig
jesconstantine: I would suggest moving this conditional block inside of the parent component `section.ma__page-banner` similar to how we see [@molecules/header-tags included on @organisms/by-template/page-header](https://github.com/massgov/mayflower/blob/master/styleguide/source/_patterns/03-organisms/by-template/page-header.twig#L6).  What do you think @labbydev @legostud ?

jesconstantine:
@legostud @gleroux02 let's talk about the architecture needed for this ticket and what is proposed in Jonathan's review comment so we can provide a path forward for @labbydev 

It seemed logical to me to add the `@molecules/header-tags` as a child pattern to whichever `@organism` was responsible for rendering the page title and other "pre-content" of the page, which would mean adding it to `@organisms/by-template/page-banner` as it already exists on `@organisms/by-template/page-header`.  The one that seems tricky, however, is `@organisms/by-template/illustrated-header` which includes `@organisms/by-template/page-header`...

The reason why it seemed logical to me is because in order to render, this pattern needs page-level data (i.e. relationship information--not hierarchical) and not site-level data.

What do people think?

gender404:
if it needs page-level data, then it might not be best to add it to the header.  I would then suggest we make it a stand alone Pattern that can be placed at the top of the page above the various Page Headers/Banners.  Similar to how we were doing breadcrumbs.

styleguide/source/_patterns/03-organisms/by-template/page-banner.twig
gender404: It can't be added there, because the background image is applied to the parent class so we would have to do a lot more work to tweak the Page Banner.  I think making it a stand alone pattern placed above the Page Banner would be a better approach.

jesconstantine: Good call @legostud that makes sense to me.

styleguide/source/_patterns/03-organisms/by-template/header-tags.json
```diff
@@ -0,0 +1,37 @@
...
+      "icon": "@atoms/05-icons/svg-twitter.twig",
+      "linkType": "twitter"
+    },{
+      "altText": "Follow us on LinkedIn",
+      "href": "#",
+      "icon": "@atoms/05-icons/svg-linkedin.twig",
+      "linkType": "linkedin"
+    }]
+  }
+}
```
gender404: Channeling Greg here, you're missing the extra line ending on your new files.

styleguide/source/_patterns/03-organisms/by-template/header-tags.json
```diff
@@ -0,0 +1,37 @@
+{
+  "headerTags": {
```
gender404: The data object for all Patterns in Mayflower, excluding pages and templates, should be contained in a single object that matches the files name.  This of course means your going to have { headerTags: { headerTags: { } } } which tells me this new pattern is named incorrectly.

Example: sample-pattern.json
~~~
{
  samplePattern {
    headerTags { ... }
    socialLinks { ... }
  }
}
~~~

styleguide/source/_patterns/03-organisms/by-template/header-tags.twig
```diff
@@ -0,0 +1,13 @@
+
+{% if headerTags.taxonomyTerms or socialLinks.items %}
+  <div class="ma__header-tags__container">
```
gender404: Incorrect BEM syntax used here.  I see a child class that isn't within a parent class.

styleguide/source/assets/scss/03-organisms/_header-tags.scss
```diff
@@ -0,0 +1,33 @@
+.ma__header-tags__container {
```
gender404: Something is obviously wrong if the outer wrapper is a child class.  Also since this pattern is called header-tags, the parent class should be .ma__header-tags { ... }.  Filename determines the class name to keep everything in sync and help avoid having multiple patterns using to same parent class.

Using .ma__header-tags isn't possible here since we are already using that elsewhere.

styleguide/source/assets/scss/03-organisms/_header-tags.scss
```diff
@@ -0,0 +1,33 @@
...
+  @include ma-container;
+  clear: both;
```
gender404: I don't think we'll need the clear both here.

styleguide/source/assets/scss/index.scss
```diff
@@ -119,6 +119,7 @@
 @import "03-organisms/form-requirements";
 @import "03-organisms/header";
 @import "03-organisms/header-alert";
+@import "03-organisms/header-tags";
```
gender404: I'm pretty sure there are theme SCSS styles (border-color) for this pattern.

gender404 (changes requested):
There is an overall issue with this new Pattern, you have used the same name for this pattern as an existing Pattern (header-tags) that is not allowed except for pages.

Simplest solution is to rename this Pattern, but I could also see us renaming the molecule to Relationship Links.

labbydev:
@legostud @jesconstantine take a look at this update and LMK how it looks to you.

jesconstantine:
@legostud @labbydev Does it make sense to call the pattern something like Content Eyebrow?

So `@organisms/by-template/content-eyebrow` would include `@molecules/header-tags` and `@molecules/social-links`.

Data structure would be:
```
{
  contentEyebrow: {
    headerTags: {},
    socialLinks: {}
  }
}
```

And we'd still remove `headerTags` from `pageHeader` -- can we do this in a minor version?

gender404:
Creating a new pattern = minor release.  
Removing code from the twig and data object for Page Header = minor release.  
Removing the CSS from page-header.scss = Major release (breaks pre-existing markup).

A minor release needs to work with pre-existing markup and data from the current Major release (5).  Adding the new pattern is easy, but modifying the Page Header will take some care to keep it backward compatible.  Technically, we don't need to modify page-header at all since the header tags are optional and can just be turned off by not passing any data.  I would still request it be removed, but that work could be done as part of a future 6.0 release.  

My suggestion is to do two PRs.  One to create the new pattern and update the pages in Mayflower (minor release) and one to remove the header tags and social links from all the Page Header files (major release).

The major release PR would be merged into a new Release-6.0 branch

jesconstantine:
If @labbydev creates the PR that you mention (creating the new pattern) for a minor release for this ticket; then @labbydev can implement that new pattern on pages that need the `headerTags` but currently use the `pageBanner` and leave the pages that use the `pageHeader` to send `headerTags` as is until we implement the future PR / major release to remove the `headerTags` + `socialLinks` from `pageHeader` and updates those pages that include `pageHeader` to use the new pattern.  I think that's out of scope for this ticket for Lauren, but Iet's at least create the ticket and ear mark it for future refactoring.

How does that sound for a path forward for everyone?

gender404:
sounds fine with me.

labbydev:
@jesconstantine @legostud sounds good to me, I just want to confirm the action items:
* [ ]  Create new ticket to remove `headerTags` from `pageHeader` 
   _@jesconstantine is this something you're going to do or do you want me to do this?_
* [x] Create a new pattern for the `headerTags` 
   _`contentEyebrow` sounds promising_
* [x] Put the styles back in `page-header.scss` to prevent backward breaking changes
* [x] If a pattern is already using `pageHeader` leave it as is (with `headerTags` and `socialLinks` included)
* [x] Additional scss fixes pointed out by @legostud (see previous comments)

Does this seem correct?

jesconstantine:
@labbydev This sounds right to me as well.  I'll create the ticket mentioned in your first bullet.

labbydev:
Thanks, sounds like a plan

styleguide/source/assets/scss/index.scss
labbydev: @legostud I couldn't find where the border color for this is being defined, do you know where that is?

labbydev:
@jesconstantine I have added the new organism and reverted the breaking changes with `pageHeader` would you like to me to add the `contentEyebrow` to any of the other patterns?

styleguide/source/_patterns/03-organisms/by-template/content-eyebrow.json
```diff
@@ -0,0 +1,39 @@
+{
+  "contentEyebrow": {
```
gender404, line 2: perfect

styleguide/source/_patterns/03-organisms/by-template/content-eyebrow.twig
```diff
@@ -0,0 +1,14 @@
+<section class="ma__content-eyebrow">
+  {% if contentEyebrow.headerTags.taxonomyTerms or contentEyebrow.socialLinks.items %}
```
gender404: I would put this outside of the section tag to remove all the html if there aren't any items to render.

styleguide/source/_patterns/03-organisms/by-template/content-eyebrow.twig
```diff
@@ -0,0 +1,14 @@
...
+        {% set headerTags = contentEyebrow.headerTags %}
+        {% include "@molecules/header-tags.twig" %}
+      {% endif %}
+      {% if contentEyebrow.socialLinks.items %}
+        {% set socialLinks = contentEyebrow.socialLinks %}
+        {% include "@molecules/social-links.twig" %}
+      {% endif %}
+    </div>
+  {% endif %}
+</section>
```
gender404: line ending.

gender404 (changes requested):
Almost there.

Missing the border color still:
image

styleguide/source/_patterns/03-organisms/by-template/content-eyebrow.json
jesconstantine: @legostud based on [our discussion re: `by-template` vs `by-author`](https://github.com/massgov/mayflower/pull/527#discussion_r131629182), should this pattern live under `by-author` instead?

jesconstantine (commented):
@labbydev this is great work - thanks!  Based on your Drupal ticket, we should see "relationship indicators" on the following Mayflower pages: 
- "Detail for service, how-to, location" page - done in Mayflower
- "Service" - let's add the `content-eyebrow` to `@templates/01-content-types/services` in the `preContent` block before the include for `pageBanner` and populate with data for both `@tempaltes/01-content-types-services` and `@pages/service`
- "Guide" - this one is tricky because it uses `illustrated-header` which includes `page-header` but @legostud and I [determined that the design doesn't support header-tags](https://github.com/massgov/mayflower/pull/524#discussion_r131552721).  I think we already have the hook to add it to `@templates/stacked-row-template` via the `preContent` block; so let's add and populate`content-eyebrow` on the guide page, by including it before the illustrated header -- what do you think, @legostud ?
- "Location" - uses `page-header` so we should be able to just populate `pageHeader.headerTags` in `@pages/location-general-content.json`
- "Topic" - uses 'page-banner' so I think we need to include `content-eyebrow` before `include "@organisms/by-template/page-banner.twig"` in `@pages/topic.twig` and add the `content-eyebrow` data structure as the first property to  `@pages/topic.json`

gender404:
The direction I received from Sarena and Ned was that every page should have relationship indicators (except the homepage I guess).  Which means any page that doesn't start with a Page Header should have this new Content Eyebrow added.

As to the Illustrated Header, the Page Header tags will not work properly in that configuration so it does need the content eyebrow.

styleguide/source/_patterns/03-organisms/by-template/content-eyebrow.json
gender404: If it was in the by-author area I would expect the user to be able to place this pattern anywhere on the page (in one of the various placeholders like Page Content or Right Rail).  This new pattern is only designed to be use in one location (Pre-Conntent) and should be at the top of that location so it has very strict placement requirements.  These requirements make it more suitable to being built into a template.

A lot of CMS systems I work on (Sitecore and AEM) allow users to drag patterns into placeholders on a page to extend the content shown there.  Using content types instead make this concept of "built in a template" a little confusing, but I'm not a better way to explain it.

labbydev:
@legostud where is that screenshot coming from? I don't see that code in `styleguide/source/assets/scss/03-organisms/_page-header.scss`? Could you possibly link me to the line in github? I'm confused since the code is basically the same between `_page-header.scss` and `_content-eyebrow.scss` what I'm missing when it comes to the border color.

labbydev:
@legostud @jesconstantine RE:Illustrated header not getting the content eyebrow, since guide pages use the illustrated header, but also require relationship indicators, how should that be addressed?

labbydev:
Should we have an option for border or no border? Seems awkward to see the border with the banner.
screen shot 2017-08-14 at 1 26 47 pm

gender404:
1. Most patterns in Mayflower have two Scss files associated with it.  The first is do the base styling and the second is to apply the Mayflower theme.  You'll see a folder in scss/06-theme. https://github.com/massgov/mayflower/blob/dev/styleguide/source/assets/scss/06-theme/03-organisms/_page-header.scss
2. Please add the new Content Eyebrow above the illustrated header on the guide page (pre-content block)
3. I agree the border looks odd there.  Feel free to add a new `hideBorder` boolean variable and `.ma__content-eyebrow--without-border` class to set `&--without-border { border: none; }` in the scss.

EDIT: that css selector isn't correct.

jesconstantine:
@labbydev @legostud for this PR let's stick with the list of page types from my [earlier comment](https://github.com/massgov/mayflower/pull/539#pullrequestreview-55965628) -- with the exception of the org page, the majority of the other page types seem to have them already anyway.

styleguide/source/_patterns/03-organisms/by-template/content-eyebrow.twig
```diff
@@ -1,5 +1,5 @@
 {% if contentEyebrow.headerTags.taxonomyTerms or contentEyebrow.socialLinks.items %}
-  <section class="ma__content-eyebrow">
+  <section class="ma__content-eyebrow{{contentEyebrow.hideBorder ? '--borderless' : ''}}">
```
gender404: modifier classes `--something` are always pair with a parent class when using BEM syntax.  This way the modifier only contains the CSS need to override or change the default styling (no duplication)

`ma__content-eyebrow ma__content-eyebrow--borderless`

styleguide/source/assets/scss/03-organisms/_content-eyebrow.scss
```diff
@@ -35,4 +35,14 @@
       }
     }
   }
+
+  &--borderless {
+    @include ma-container;
+    @include clearfix;
```
gender404: these two includes shouldn't be needed once the you fix my other comment.

gender404: it will also need to be `&--borderless &__tags {` to achieve the proper specificity

gender404 (changes requested):
almost there.  Nice job.

gender404 (approved):

jesconstantine (approved):
Thank you @labbydev - this is great work!

PR #541, opened 2017-08-08
Dp 4297 footnote fix

gleroux02:
Footnotes have been simplified and there might not always be all of the data needed to display the footnote anchor. If there are no numbered footnotes hide the anchor.

### DIFF
--- a/styleguide/source/_patterns/02-molecules/footnote.twig
+++ b/styleguide/source/_patterns/02-molecules/footnote.twig
@@ -1,5 +1,5 @@
 <div class="ma__footnote js-footnote" tabindex="-1">
-  <a class="ma__footnote__link js-footnote-link" id="{{ footnote.id }}" href="{{ footnote.target }}">{{ footnote.number }}</a>
+  {% if footnote.number %}<a class="ma__footnote__link js-footnote-link" id="{{ footnote.id }}" href="{{ footnote.target }}">{{ footnote.number }}</a>{% endif %}
   <div class="ma__footnote__rich-text js-footnote-message">
     {% set richText = footnote.richText %}
     {% include "@organisms/by-author/rich-text.twig" %}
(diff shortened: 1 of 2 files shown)


gender404:
I'm confused by this change.  Not having a number seems to defeat the purpose of footnotes since we don't know what this relates to which makes me wonder if the request was actually to remove the link from number instead of removing the number.

Outside of that there are a couple issues here.  
1. The `js-footnote` class needs to be removed to prevent the JavaScript from being processed on this note.
2. There is extra padding on the left side of the footnote where the number used to be.  Should that still be there when the number is removed?

styleguide/source/_patterns/02-molecules/footnote.twig
```diff
@@ -1,5 +1,10 @@
-<div class="ma__footnote js-footnote" tabindex="-1">
-  {% if footnote.number %}<a class="ma__footnote__link js-footnote-link" id="{{ footnote.id }}" href="{{ footnote.target }}">{{ footnote.number }}</a>{% endif %}
+{# Option to remove js functionality and hide the footnote link (this is an implementation workaround). #}
+{% set footnoteJs = footnote.number ? '' : 'js-footnote' %}
```
gender404: this is backwards.

gender404 (changes requested):

styleguide/source/_patterns/02-molecules/footnote.twig
jesconstantine: Ha - I was just testing YOU. :) I'll fix it, thanks.

jesconstantine: Those tricky ternaries.  Fixed in https://github.com/massgov/mayflower/pull/541/commits/812aa25546acd5b6cebc3e9f3509e41f99a389ec

gender404 (approved):

PR #543, opened 2017-08-09
DP-1900 - Changed decorative links within contact groups to display inline.

smurrayatwork:
**Description:** 
I used the contact-group.css override that already existed to add a new CSS rule. This is to make decorative links inline **only** when they're within the contact group. Without this, Firefox renders these links with a hyphenated (or sometimes not) break.

See the following in Firefox (version 52.2.1 ESR):

- http://mayflower.digital.mass.gov/?p=pages-location-general-content (link for "Directions")
- https://pilot.mass.gov/locations/southbridge-rmv-service-center (link for "Directions" and "RMV contact form").

**Jira:** 
https://jira.state.ma.us/browse/DP-1900

**To Test:**
- [ ] Browse to any of the sample links above. Verify that the links for "Directions" or "RMV contact form" no longer word break as described.

**Post Deploy Steps:**
N/A

**Relevant Screenshots/GIFs:**
mass gov_bluehills_reservation_directionslink_onfullbrowserpageload_linktextcuttoff_viapositioning

**Additional Notes:**
N/A

### DIFF
--- a/styleguide/source/assets/scss/02-molecules/_contact-group.scss
+++ b/styleguide/source/assets/scss/02-molecules/_contact-group.scss
@@ -60,7 +60,7 @@
 
   .ma__content-link,
   .ma__decorative-link {
-    display: inline-block;
+    display: inline;
     width: auto;
 
     a {
--- a/styleguide/source/assets/scss/02-molecules/_contact-group.scss
+++ b/styleguide/source/assets/scss/02-molecules/_contact-group.scss
@@ -60,8 +60,7 @@
 
   .ma__content-link,
   .ma__decorative-link {
-    display: inline;
-    width: auto;
+    display: inline-block;
 
     a {
       hyphens: auto;


styleguide/source/assets/scss/02-molecules/_contact-group.scss
```diff
@@ -60,7 +60,7 @@
 
   .ma__content-link,
   .ma__decorative-link {
-    display: inline-block;
+    display: inline;
     width: auto;
```
gender404: I'm pretty sure you could have just removed the width: auto instead.  I'm hesitant to change the value to 'inline' since part of the solution to keep the arrow from wrapping early relies on that element being inline-block;

smurrayatwork: @legostud I had tried removing this without display: inline and it didn't resolve the issue on firefox.

Actually, trying this again, and you're right. I'll update this to remove width: auto instead.

PR #552, opened 2017-08-22
Atomic design

gender404:
Documentation / rules about Atomic Design for Mayflower.

### DIFF
--- a/docs/atomic-design.md
+++ b/docs/atomic-design.md
@@ -8,19 +8,18 @@ You can find an excellent description of Atomic Design on Brad Frost's [blog](ht
 We are following his guide fairly closely, but I differ when it comes to Molecules and Organisms. He sees then as being defined based on their complexity, but I found in practice that it was difficult to make that distinction.  I like to think of Molecules as larger Atoms (components with custom styling or additional content) and Organisms as being smaller templates (component level)  
 
 ### Atoms (an "indivisible unit")
-An Atom is the smallest logically stylized pattern in Atomic Design.  Dividing an Atom into smaller parts would result in an unstable pattern that would be dependent on another pattern.  As such Atoms would never include other patterns.
+An Atom is the smallest possible pattern in Atomic Design.  They are the building blocks to create more complex Patterns.  In most cases, a simple html element (`a`, `h1`, `button`, etc...) is considered an Atom.  More complex html elements (`table`, `ul`, `figure`, etc...) are also considered Atoms, because the individual parts (`li`,`td`,`figcaption`) cannot be used on there own.
 
-An Atom can be a single HTML element (`a`, `h1`, etc...) or a single complex HTML5 elment that has nested dependent children (`table`, `ul`, `figure`, etc...).  It could also be two elements that would never be used seperately like an input with a label.
+There can be cases where a simple html element is too small to be considered an Atom.  If your design creates a dependency between two elements, then they should be combined into a single Atom.  For example, an `input` is always supposed to be paired with a `label` for accesibiltiy.  In this case, the `input` element alone would be too small to be an Atom.  Instead, you would pair the `input` with a `label` to create the Input Atom.
 
-Since Atoms are so small they are rarely if ever used directly on a page.  Instead they are used as building blocks to create more complex Patterns.
 
 #### Quick guide:
-* Could be a single simple (`a`) or complex (`table`) HTML element
-* Might contain multiple HTML elements, if one element cannot be rendered without the other
+* A simple html element (`a`, `h1`, `button`, etc...) or complex (`table`, `ul`, `figure`, etc...) HTML element
+* Might pair html elements if there is a dependency
 * Almost always uses a theme SCSS file to style the element
-* Cannot be broken down into smaller logical elements
-* Usually contains content data
-* Sometimes includes an icon
+* Contains the smallest set of elements possible
+* Almost always renders content data
+* Can contain `include` statements to add an icon or another dependent Atom.
 
 #### Examples
 1. http://mayflower.digital.mass.gov/?p=atoms-button
@@ -28,12 +27,14 @@ Since Atoms are so small they are rarely if ever used directly on a page.  Inste
 2. http://mayflower.digital.mass.gov/?p=atoms-input-checkbox
     * This is an example of an atom that can't be broken down any further.  The Mayflower style guide would never use the `input type=checkbox` element by itself so this atom requires the outer span tag and corresponding label. 
 
-### Molecules - (stylized collections)
-A Molecules is a stylized collection of Atoms, Molecules, and or HTML elements.  Molecules have the greatest flexibility of all the patterns.
+
+### Molecules - (stylized collection)
+A Molecule is a stylized collection of Atoms, HTML elements and even other Molecules that are combined to create a new pattern for a simple task.  These patterns usually accomplishs a single goal like a stylized link (ie: Callout Link) or a menu of links (ie: Footer Links).  Molecules have the greatest flexibility of all the patterns so please make sure you're not building an Atom or Organism.
 
 #### Quick guide:
-* Includes other patterns or HTML elements
-* Usually contains content data
+* Includes other patterns (Atoms or Molecules) or HTML elements
+  * Currently, some patterns are incorrectly categorized so including an Organism is acceptable. (ie: Rich Text organism)
+* Can render content data or pass it on to other patterns
 * Almost always uses a theme SCSS file to style the element
 
 #### Example
@@ -45,7 +46,7 @@ http://mayflower.digital.mass.gov/?p=molecules-image-promo
 An Organism is a layout of other patterns.  HTML5 elements and CSS styling used in an Organism are for laying out the included patterns.  All content to be rendered, is passed on to the included patterns.  The only styling done for an Organism would be to help define the layout the patterns, like background-colors and borders.
 
 #### Quick guide:
-* Includes other patterns.
+* Includes other patterns (Atoms, Molecules, or Organisms).
 * All content data is rendered by included patterns.
 * CSS styles used are only for layout (flex box, float ...) or to visually define the layout (backgrounds, borders)
 * Often times it doesn't use a theme SCSS file
--- a/docs/atomic-design.md
+++ b/docs/atomic-design.md
@@ -1,34 +1,58 @@
 # Atomic Design in Mayflower
 
 ## About Atomic Design
-This design system is a way of break down a design into smaller re-usable parts.  Starting at the page level, you would look for common patterns between pages, which would be become templates.  You would then look for common components on these pages which might become Organisms or Molecules.  Organisms can then be broken down further into Molecules and Atoms or even other Organisms.  Molecules can then be broken down into Atoms or even Molecules.  Eventually your design is broken down into the smallest indivisible units giving you a re-usable palette to build new patterns from.
+This design system is a way of looking at a design as a collection of smaller re-usable parts.  Starting at the page level, you would look for common patterns between pages, which would be become templates.  You would then look for common components on these pages which might become Organisms or Molecules.  Organisms can then be broken down further into Molecules and Atoms or even other Organisms.  Molecules can then be broken down into Atoms or even Molecules.  Eventually your design is broken down into the smallest indivisible units giving you a re-usable palette to build new patterns from.
 
 As defined by [Brad Frost](http://atomicdesign.bradfrost.com/chapter-2/).
 
 
-### Atoms
-* A single HTML5 element type (includes child elements - ul -> li or figure -> figcaption)
-* Should be an "indivisible unit" (ie: Input pattern would always be used with a label, but there wouldn't be a Label pattern since it would never be used on it's own).
-* Never includes other patterns.
+### Atoms (an "indivisible unit")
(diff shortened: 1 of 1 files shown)


docs/atomic-design.md
```diff
@@ -0,0 +1,61 @@
+# Atomic Design in Mayflower
+
+## About Atomic Design
+Atomic Design is a way of viewing a web page as a collection of smaller re-usable parts.  It's very similar to building with LEGO bricks.  Each piece can be combinded with others making it quicker and easier to build web pages.
```
millsb: Technically, Atomic Design is about defining the parts of an entire design system, not just a tool for breaking down individual pages. The goal would be to encourage systematic thinking around all parts of a website (or an entire online brand presence, really) .

docs/atomic-design.md
```diff
@@ -0,0 +1,61 @@
+# Atomic Design in Mayflower
+
+## About Atomic Design
+Atomic Design is a way of viewing a web page as a collection of smaller re-usable parts.  It's very similar to building with LEGO bricks.  Each piece can be combinded with others making it quicker and easier to build web pages.
+
+The easiest approach to get stated is to a take a complete web page and try breaking it down.  Viewing a web page, you would look for areas that would be consistent between each variant of that page to help determine your Template.  From that template, you would look for  logical groups (components) which might become Organisms or Molecules.  Even these patterns might be able to broken down into smaller parts so you start looking for common individual elements (buttons, headers, lists, ...) to find your Atoms.  Eventually your web page has been broken down into the smallest indivisible units giving you a re-usable palette of pieces to build new patterns and web pages.
```
millsb: I don't want to hammer on this too much, but I think if we're doing Atomic Design right, defining the Atom/Molecule/Organism is part of the design phase, and not exclusive to development. Taking existing design and 'converting' it to Atomic feels like missing the point of Atomic. Defining the design semantics of a piece help inform usage (and re-usage) just as much as the code does. I know it wasn't the case in Mass.gov, but would love to push this idea going forward.

docs/atomic-design.md
```diff
@@ -0,0 +1,61 @@
+# Atomic Design in Mayflower
+
+## About Atomic Design
+Atomic Design is a way of viewing a web page as a collection of smaller re-usable parts.  It's very similar to building with LEGO bricks.  Each piece can be combinded with others making it quicker and easier to build web pages.
+
+The easiest approach to get stated is to a take a complete web page and try breaking it down.  Viewing a web page, you would look for areas that would be consistent between each variant of that page to help determine your Template.  From that template, you would look for  logical groups (components) which might become Organisms or Molecules.  Even these patterns might be able to broken down into smaller parts so you start looking for common individual elements (buttons, headers, lists, ...) to find your Atoms.  Eventually your web page has been broken down into the smallest indivisible units giving you a re-usable palette of pieces to build new patterns and web pages.
+
+As defined by [Brad Frost](http://atomicdesign.bradfrost.com/chapter-2/).
```
millsb: Consider leading with the link to Brad's Atomic Design docs? If we can send people to the primary source right away, we don't have to cover all the basics and spend more time talking about how this applies to mass.gov

docs/atomic-design.md
```diff
@@ -0,0 +1,61 @@
...
+
+As defined by [Brad Frost](http://atomicdesign.bradfrost.com/chapter-2/).
+
+
+### Atoms (an "indivisible unit")
+An Atom is the smallest logical stylized pattern in Atomic Design.  Dividing an Atom into smaller parts would result in an unstable pattern that would be dependent on another pattern.  As such Atoms would never include other patterns.
+
+An Atom can be a single HTML element (`a`, `h1`, etc...) or a single complex HTML5 elment that has nested dependent children (`table`, `ul`, `figure`, etc...).  It could also be two elements that would never be used seperately like an input with a label.
+
+Since Atoms are so small they are rarely if ever used directly on a page.  Instead they are used as building blocks to create the more complex Patterns.
```
millsb: You make a good distinction here, as Atoms are the hardest bit to grasp.

docs/atomic-design.md
```diff
@@ -0,0 +1,61 @@
...
+Molecules have the greatest flexibility of all the patterns so you should use caution when using HTML elements to make sure they shouldn't be an included Atoms instead.
+
+#### Quick guide:
+* Includes other patterns or HTML elements
+* Includes content data
+* Usually has a theme SCSS file
+ 
+
+### Organisms
+Like Molecules, Organisms are built by including other patterns to create a more complex pattern.  The difference is that the only HTML5 elements and CSS styling used is for laying out this patterns and all content is passed on to these patterns.  The only styles used in an Organism would be to layout the patterns and include some theming css like background-colors and borders to help define the layout.
```
millsb: A good rule of thumb.

millsb:
This is a good start. I think you've covered a lot of the How, but might need some more Why. Have you considered linking out to direct examples of Atoms/Molecules/Organisms in Mass.gov and explaining why we made a distinction there, and how we intend them to be used?

docs/atomic-design.md
gender404: yeah, changing the order might be a good idea.  I could probably even quote something from his blog.  The issues I had with his article were around the definitions between types, but I've addressed that below so his article should be pretty good no that I think about it.

docs/atomic-design.md
gender404: The issue I see is that it seems like it easier to understand how to make the Atoms once you know what the page should look like.  As a way of explaining it from the other direction.  I think you're right about the previous paragraph though in that it should describe it going from Atom to Organism and then this could potentially go the other way as a comparison.

gender404:
@millsb - I made some updates to the copy to make the intro more design and added examples to the Atoms, Molecules, and Organisms.  Let me know if this reads better.

docs/atomic-design.md
```diff
@@ -0,0 +1,70 @@
...
+* CSS styles used are only for layout (flex box, float ...) or to visually define the layout (backgrounds, borders)
+* Often times it doesn't use a theme SCSS file
+* Data used by the Organism is used to control the layout
+
+#### Example
+1. http://mayflower.digital.mass.gov/?p=organisms-event-listing
+    * This is an example of an organism that includes other patterns, passes all content data to another pattern, has theme based css related to layout applied, and has a data variable to switch between a list and grid views.
+
+### Templates
+* Template patterns serve the same purpose of an Organism, but they are used to represent a complete page.  Since we're working at the page level, templates should contain Twig Blocks as placeholders to allow pages to add patterns or to override default patterns contained within the Template pattern.
```
jesconstantine: I'd like it if templates did some content modeling as well - i.e. Page Title (125 characters max)

docs/atomic-design.md
```diff
@@ -0,0 +1,70 @@
+# Atomic Design in Mayflower
+
+## About Atomic Design
+The Mayflower style library is built using Atomic Design.  You can find an excellent description of Atomic Design on Brad Frost's [blog](http://atomicdesign.bradfrost.com/chapter-2/).
```
jesconstantine: Especially since we're linking to it, I think we need to call out how we align and differentiate with Atomic Design here.

docs/atomic-design.md
```diff
@@ -0,0 +1,70 @@
...
+* Template patterns serve the same purpose of an Organism, but they are used to represent a complete page.  Since we're working at the page level, templates should contain Twig Blocks as placeholders to allow pages to add patterns or to override default patterns contained within the Template pattern.
+
+
+#### Quick guide:
+* Similar to Organisms, but at the page level
+* Provides the base structure for a page
+* Contains Twig Blocks to include optional patterns or to override default patterns
+
+### Pages
+Pages are used as an example of what the final Page for the site should look like.  Pages always start by Extending an existing template.
```
jesconstantine: Vs templates which model content, pages should show "real" content

jesconstantine (commented):
This is off to a great start @legostud !

docs/atomic-design.md
gender404: so ... something like Template can manipulate data before rendering included patterns?

docs/atomic-design.md
```diff
@@ -0,0 +1,90 @@
...
+The Mayflower style library is built using Atomic Design. The way I tend to think of building a site using Atomic Design is to see it as a LEGO set where your combining a collection of small parts  to make a variety of different models.  Atomic Design is similar in nature, where you define a collection of small html parts called Atoms and then use those to build a variety of web pages. 
+
+You can find an excellent description of Atomic Design on Brad Frost's [blog](http://atomicdesign.bradfrost.com/chapter-2/). 
+
+We are following his guide fairly closely, but I differ when it comes to Molecules and Organisms. He sees then as being defined based on their complexity, but I found in practice that it was difficult to make that distinction.  I like to think of Molecules as larger Atoms (components with custom styling or additional content) and Organisms as being smaller templates (component level)  
+
+### Atoms (an "indivisible unit")
+An Atom is the smallest logically stylized pattern in Atomic Design.  Dividing an Atom into smaller parts would result in an unstable pattern that would be dependent on another pattern.  As such Atoms would never include other patterns.
```
isaacchansky: I'm confused on the phrase "an unstable pattern that would be dependent on another pattern". Is this saying that anything smaller than an Atom is unstable and would depend on the Atom?

docs/atomic-design.md
```diff
@@ -0,0 +1,90 @@
...
+
+An Atom can be a single HTML element (`a`, `h1`, etc...) or a single complex HTML5 elment that has nested dependent children (`table`, `ul`, `figure`, etc...).  It could also be two elements that would never be used seperately like an input with a label.
```
isaacchansky: Why the distinction between HTML and HTML5 here?

docs/atomic-design.md
```diff
@@ -0,0 +1,90 @@
...
+1. http://mayflower.digital.mass.gov/?p=atoms-button
+    * This is an example of a single html element => `button` or `a` tag.
+2. http://mayflower.digital.mass.gov/?p=atoms-input-checkbox
+    * This is an example of an atom that can't be broken down any further.  The Mayflower style guide would never use the `input type=checkbox` element by itself so this atom requires the outer span tag and corresponding label. 
+
+### Molecules - (stylized collections)
+A Molecules is a stylized collection of Atoms, Molecules, and or HTML elements.  Molecules have the greatest flexibility of all the patterns.
+
+#### Quick guide:
+* Includes other patterns or HTML elements
```
isaacchansky: When you say patterns does that only mean Atoms? Do any molecules _not_ contain Atoms?

docs/atomic-design.md
```diff
@@ -0,0 +1,90 @@
...
+#### Example
+http://mayflower.digital.mass.gov/?p=molecules-image-promo
+* This is an example of a molecule that includes other patterns, renders content data and has some theme based css applied.
+
+
+### Organisms - (layout of patterns)
+An Organism is a layout of other patterns.  HTML5 elements and CSS styling used in an Organism are for laying out the included patterns.  All content to be rendered, is passed on to the included patterns.  The only styling done for an Organism would be to help define the layout the patterns, like background-colors and borders.
+
+#### Quick guide:
+* Includes other patterns.
```
isaacchansky: Are the patterns specifically only Atoms/Molecules?

docs/atomic-design.md
gender404: that's just a typo.

docs/atomic-design.md
gender404: I'll see if I can find a better phrasing for this.

docs/atomic-design.md
gender404: Patterns means anything from Atoms to Organisms.  (I can add that though)

I guess, it's possible for a Molecule to not include another pattern, but I would find that very odd considering the number and variety of Patterns we have, but I'll change it to 'Can'

docs/atomic-design.md
gender404: Again Atoms to Organisms, but I'll expand the statement to make it more obvious.

gender404:
The biggest obstacle with Atomic Design is trying to define Molecules.  When you read Brad Frosts blog he states the only difference between Organisms and Molecules is complexity, which is a hard thing to measure.  Is complexity three or four included patterns.  Is it the type of pattern, listing vs image promo.  Maybe Organisms really should be templates.  

I could say it only includes Atoms, but that doesn't work for a lot of simple patterns like image promo which has a Rich Text (molecule or organism).

Although, I could try to define complexity as being a single task.  This pattern does one thing.  And add that as an additional requirement.

And the biggest issue is that Mayflower was originally built with Organisms equaling a component (something usable on a page by itself) whereas Molecules were built as a sub assembly of atoms to be consumed by an Organism before being used on the page.  That concept though fell apart pretty quick with the Rich Text and the Colored Heading patterns which are both consumable and usable.

gender404:
Updates have been made

jesconstantine:
@isaacchansky any thoughts on the best place to hook into this doc?  

I'm thinking both from the repo readme and also from the contrib docs, maybe before / after working with pattern lab: https://github.com/massgov/mayflower/blob/dev/.github/CONTRIBUTING.md#working-with-pattern-lab

~Oooo or maybe from the work in~ Nevermind - @legostud already added it in his fabulous technical contrib docs: https://github.com/massgov/mayflower/pull/607/files

Let's get both of these PRs merged - what do you think?

jesconstantine (approved):
Thanks @legostud this rocks!

PR #558, opened 2017-08-23
DP-4581 - Init event listing

nstriedinger:
**Styleguide:**
http://mayflower.digital.mass.gov/?p=pages-event-listing

### DIFF
--- a/styleguide/source/_patterns/03-organisms/by-author/event-listing.twig
+++ b/styleguide/source/_patterns/03-organisms/by-author/event-listing.twig
@@ -1,4 +1,5 @@
 {% set gridClass = eventListing.grid ? 'ma__event-listing--grid' : '' %}
+{% set maxItems = eventListing.maxItems ?: eventListing.events.length %}
 
 <section class="ma__event-listing {{ gridClass }}">
   <div class="ma__event-listing__container">
@@ -10,13 +11,7 @@
       {% set sidebarHeading = eventListing.sidebarHeading %}
       {% include "@atoms/04-headings/sidebar-heading.twig" %}
     {% endif %}
-    <ul class="ma__event-listing__items">
-      {% for eventTeaser in eventListing.events %}
-        <li class="ma__event-listing__item">
-          {% include "@molecules/event-teaser.twig" %}
-        </li>
-      {% endfor %}
-    </ul>
+    <ul class="ma__event-listing__items"></ul>
     {% if eventListing.more %}
       <div class="ma__event-listing__more">
         {% set link = eventListing.more %}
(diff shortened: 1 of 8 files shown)


styleguide/source/_patterns/03-organisms/by-author/event-listing.twig
```diff
@@ -10,13 +11,7 @@
...
       {% include "@atoms/04-headings/sidebar-heading.twig" %}
     {% endif %}
-    <ul class="ma__event-listing__items">
-      {% for eventTeaser in eventListing.events %}
-        <li class="ma__event-listing__item">
-          {% include "@molecules/event-teaser.twig" %}
-        </li>
-      {% endfor %}
-    </ul>
+    <ul class="ma__event-listing__items"></ul>
```
gender404, line 14: I'm assuming your planning on having JS render this content anywhere it gets used.  It would be better to leave it as is and just have JS regenerate the list on the Event Listing page.

styleguide/source/assets/js/modules/eventListing.js
```diff
@@ -0,0 +1,126 @@
+import sticky from "../helpers/sticky.js";
+import getTemplate from "../helpers/getHandlebarTemplate.js";
+
+export default function (window,document,$,undefined) {
+  let listing = $('.ma__event-listing__items');
+  let submitButton = $('.ma__event-filters__submit');
+  let resultsHeader = $('.ma__results-heading__title');
```
gender404, line 7: CSS classes should not be used in a JS file.  Use `js-event-listing-items` style classes instead to keep CSS styles and JS functionality separated.

styleguide/source/assets/js/modules/eventListing.js
```diff
@@ -0,0 +1,126 @@
+import sticky from "../helpers/sticky.js";
+import getTemplate from "../helpers/getHandlebarTemplate.js";
+
+export default function (window,document,$,undefined) {
```
gender404, line 4: I feel like this code isn't isolated (namespaced) enough to prevent some of it from running on the location listing page.

I would have expected a $('js-event-listing-page').each(function(){ ... }); used to initialize and start the code.

styleguide/source/assets/js/templates/eventListingRow.html
```diff
@@ -0,0 +1,37 @@
+<li class="ma__event-listing__item">
+  <section class="ma__event-teaser">
```
gender404, line 2: I'm concerned with the `li` being included here, because it makes it less re-usable.  Maybe there is another way.

gender404 (changes requested):
It's a decent start, but needs some help.  I would suggest adding a `js` class to the page itself to start the event-listing.js code.

jesconstantine:
Close in favor of #564 (WIP)

PR #559, opened 2017-08-23
Dp 3631 event colored heading

robyngreen:
Adds support for colored heading to eventListing displays, per discussion and requirements in DP-3631

This also adds elseif, as I'm not sure the template was intended to support both displays potentially at the same time.

### DIFF
--- a/styleguide/source/_patterns/03-organisms/by-author/event-listing.twig
+++ b/styleguide/source/_patterns/03-organisms/by-author/event-listing.twig
@@ -2,15 +2,15 @@
 
 <section class="ma__event-listing {{ gridClass }}">
   <div class="ma__event-listing__container">
-    {% if eventListing.coloredHeading %}
-      {% set coloredHeading = eventListing.coloredHeading %}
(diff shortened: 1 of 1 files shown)


gender404 (changes requested):
This change isn't needed.  The Location and How To pages use content types that are coded in Mayflower to convert any Sidebar or Comp Heading to render as a Colored Heading.  This is accomplished by added a single class to the Right Rail `sidebar sidebar--colored`

I'm guessing your not using the most recent Location Content Type:  
http://localhost:3000/?p=templates-location

robyngreen:
Ah @legostud that's awesome and great to hear. It actually looks like the problem then is rooted more in the node--location.html.twig, which if it needs to behave like you said with node--how-to-page.html.twig, then it needs to be updated as it's still using `{% block sidebar %}` where howto uses `<aside class="sidebar sidebar--colored">`

robyngreen:
This can be solved by adding the `<aside class="sidebar sidebar--colored">` to the location template, as seen in other node page templates. The change is unnecessary.

PR #565, opened 2017-08-24
Updating Gulp Packages to work with npm 5.2.0

gender404:
Note: npm 5.3.0 didn't work with node-sass on Windows 10.

I recently installed node v6.11.2 and npm v5.2.0 which caused the node-normalize-scss package to fail


jesconstantine (commented):
@legostud I deleted `styleguide/npm_modules`, ran `npm install`, then touched scss file, a .json file, and a .twig file and it compiled and reloaded successfully after all of those.  Anything else I should test for this?

jesconstantine (approved):

PR #566, opened 2017-08-24
DP-4730 Render event end date in date summary on event page

jesconstantine:
This PR adds a filter to `@molecules/event-teaser` around `eventTeaser.date.summary` to support multiple lines.  This will allow a event teaser date summary to show both start and end date.  It also adds to variants to the event page - one showing event start and end date with separate times; one showing event start and end date with the same time window.

JIRA ticket: https://jira.state.ma.us/browse/DP-4730

### To Test

- [ ] Pull down branch
- [ ] `cd styleguide` && `gulp`
- [ ] Browse to http://localhost:3000/?p=pages-event-end-date
- [ ] Confirm it meets visual design requirements from ticket (screenshot attached below)

### Screenshot
image


gender404 (approved):
Seems odd to create an entire page variant just to show the these differences.  Not sure a what a better solution would be.  Page Header variants could work.

I'll approve this anyways.

jesconstantine:
@legostud agreed - I think I opted for page since the only real variant was content and not a variable change.  Make sense?

gender404:
that makes sense

PR #573, opened 2017-08-29
Release 5.7.0

mrossi113:
### Added

- DP-4730 - Users will now notice there is now an option to render an [Event page with an end date](http://mayflower.digital.mass.gov/?p=pages-event-end-date). ([PR #566](https://github.com/massgov/mayflower/pull/566))
- DP-4881 - The [Press release page](http://mayflower.digital.mass.gov/?p=pages-press-release) now renders with (optional) subtitle content. ([PR #556](https://github.com/massgov/mayflower/pull/556))
- DP-3235 - Screen reader users are now informed of the [download link](http://mayflower.digital.mass.gov/?p=molecules-download-link) file size. ([PR #522](https://github.com/massgov/mayflower/pull/522))
- DP-3234 - Screen reader users are now informed of a page's category (i.e. Decision, Regulation) in [the context of the page title](http://mayflower.digital.mass.gov/?p=organisms-page-header).  (See also [`@organisms/by-template/illustrated-header`](http://mayflower.digital.mass.gov/?p=organisms-illustrated-header), [PR #509](https://github.com/massgov/mayflower/pull/509))
- DP-3232, 1369 - Screen reader users can now associate key actions link content with their respective heading content.  Also, there is now a [variant to the Comp heading](http://mayflower.digital.mass.gov/?p=atoms-comp-heading-as-sidebar-heading) which renders styled as a Sidebar Heading. ([PR #518](https://github.com/massgov/mayflower/pull/518))
- DP-5175 - A "more" link will now render on different sections of [Service](http://mayflower.digital.mass.gov/?p=pages-service) and [Service Detail](http://mayflower.digital.mass.gov/?p=pages-detail-for-service-howto-location) pages. ([PRs #570](https://github.com/massgov/mayflower/pull/570), [#551](https://github.com/massgov/mayflower/pull/551))
- DP-4591 - A "learn more" link has been added to the [contact us molecule](http://mayflower.digital.mass.gov/?p=molecules-contact-us-collapsed-with-more-link), as seen on [How-To pages](http://mayflower.digital.mass.gov/?p=pages-howto) which can provide a link to learn more about an organization referenced by a contact. ([PR #542](https://github.com/massgov/mayflower/pull/542))
[...]

PR #574, opened 2017-08-30
DP-4211 - Video - making updates backward compatible.

gender404:
Added a .ma__video--new class to the video pattern so old styles can be applied if it has not been added.  Eventually we can remove this class with the next Major release.

### DIFF
--- a/styleguide/source/assets/scss/01-atoms/_video.scss
+++ b/styleguide/source/assets/scss/01-atoms/_video.scss
@@ -66,4 +66,9 @@
       margin-right: calc(50vw - (#{$max-width}/2) + 40px);
     }
   }
+
+  .ma__content-link,
+  .ma__decorative-link {
+    margin-top: 12px;
+  }
 }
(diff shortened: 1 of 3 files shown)


jesconstantine (approved):
@legostud If I am understanding it correctly, this PR is creating an opt-in (via class) accessibility presentation fix to the video atom pattern and without the opt-in the video pattern will render as it does in production today.  Thanks for the work around!

styleguide/source/assets/scss/01-atoms/_video.scss
```diff
@@ -15,11 +15,55 @@
     }
   }
 
+  // Backward Compatible - had to add the --new as a variant for the new styling
   .main-content--full .page-content > & {
-    @include ma-container();
+    @include ma-container;
   }
 
   &__link {
```
jesconstantine, line 23: See [comment in release PR](https://github.com/massgov/mayflower/pull/573/files#r136220998) about making this backwards compatible.

gender404:
It's an ugly hack, but it was the only way to make it work.  We can remove the dependency for that class in the next Major release.

styleguide/source/assets/scss/01-atoms/_video.scss
gender404: good catch

PR #575, opened 2017-08-30
DP-5342 Google Translate

gender404:
https://jira.state.ma.us/browse/DP-5342

Built a new Pattern that contains all of the functionality and styles needed for Google Translate menu.

New Pattern is called Language Menu.


styleguide/source/_patterns/03-organisms/by-template/footer.twig
```diff
@@ -28,21 +28,3 @@
...
-
-    (function() {
-      var script = document.createElement('script');
-      script.src = "//translate.google.com/translate_a/element.js?cb=googleTranslateElementInit";
-      script.async = true;
-      element = document.getElementsByTagName('head')[0];
-      element.appendChild(script);
-    })();
-  </script>
-{% endif %}
```
jesconstantine, line 48: We'll need a "migrate path" note to mention removing this from any "footer" templates which are often implementation-specific.

styleguide/source/_patterns/03-organisms/by-template/utility-nav.twig
```diff
@@ -2,12 +2,7 @@
   <ul class="ma__utility-nav__items">
     {% if googleLanguages %}
       <li class="ma__utility-nav__item">
-        <div class="ma__utility-nav__translate">
-          <div id="google_translate_element"></div>
-          <div class="ma__utility-nav__translate-icon">
-            {% include "@atoms/05-icons/svg-lat-lon-globe.twig" %}
-          </div>
-        </div>
+        {% include "@organisms/by-template/language-menu.twig" %}
```
jesconstantine, line 5: I think the only thing that makes this not backwards-compatible is that you've got an ID `google_translate_element` hardcoded in the legacy markup which works with the legacy js `googleTranslateElementInit()` but in your new js you use a different id `language-menu-0`.  So the new js would not work with the older markup.  Does that sound right?

That being said, I know for massgov we pull in the utility-nav twig so as long as we remove the old js, we should be fine...

jesconstantine (commented):
Nice work - I like the idea of keeping the JS with the new pattern itself.  I do have a couple questions re: backwards compatibility.  Let me know what you think.

Also, it'd be nice to not have to repeat the logic `{% if googleLanguages %}` in both the parent pattern and the pattern itself but I think I understand why you did it ...

styleguide/source/_patterns/03-organisms/by-template/utility-nav.twig
gender404: actually this makes it backward compatible since we don't want the new JS to work with the old markup.  If the user only updates the JS file, then the utility nav with still have the `#google_translate_element` and footer will still have the old `script` tag that populates that id.

If this code gets updated prior to the footer code, I don't think will be a major issue.  The footer code will probably throw and error when it tries to add a class to `undefined`, but that shouldn't affect any other JS code running.

PR #580, opened 2017-09-01
Make the location listing page work in IE 11

gender404:
The Location Listing page was generating an error in IE11.  JavaScript code used in the foot.twig file was trying to use ES6 javascript which is not supported.  I replaced that code with a jQuery trigger event instead.

### To test
1. browse to the location listing page in staging with IE 11: https://jesconstantine.github.io/mayflower/?p=pages-location-listing

Make sure the zip sort, tag filter, pagination, and filter clearing all work

Here's a screenshot of the location listing functioning in IE11 with:
- tag filter applied (indicated by the results heading tags)
- svgs rendering in the updated results

image

### DIFF
--- a/styleguide/source/assets/js/helpers/getElementOuterHtml.js
+++ b/styleguide/source/assets/js/helpers/getElementOuterHtml.js
@@ -1,0 +1,18 @@
+/**
+  * Get outerHTML of elements, taking care
+  * of SVG elements in IE as well.
+  *
+  * @param {Element} jquery object el
+  * @return {String}
+  */
+
+export default function (el) {
+  if (el.outerHTML) {
+    return el.outerHTML
+  }
+  else {
+    let container = document.createElement('div');
+        container.appendChild(el.cloneNode(true));
+    return container.innerHTML;
+  }
+}
--- a/styleguide/source/assets/js/helpers/getElementOuterHtml.js
+++ b/styleguide/source/assets/js/helpers/getElementOuterHtml.js
@@ -2,7 +2,7 @@
   * Get outerHTML of elements, taking care
   * of SVG elements in IE as well.
   *
-  * @param {Element} dom element el
+  * @param {Element} DOM object
   * @return {String}
   */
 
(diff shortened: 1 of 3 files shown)


jesconstantine (approved):
@legostud I don't think you can review your own PR but can you look at the files touch by my 3 commits?

jesconstantine:
@nstriedinger you'll want to make sure that your event listing PR has these updates so we don't have a regression in IE.

styleguide/source/assets/js/helpers/getElementOuterHtml.js
```diff
@@ -0,0 +1,18 @@
+/**
+  * Get outerHTML of elements, taking care
+  * of SVG elements in IE as well.
+  *
+  * @param {Element} jquery object el
+  * @return {String}
+  */
+
+export default function (el) {
```
gender404: if this takes in a jQuery object, then I would rename the argument to $el to be consistent.

jesconstantine: It takes a dom element, which we get from a jQuery selector: `$('.js-filter-by-tags').find("#" + tag).parent().siblings('svg')[0]`.  I think it would be more helpful to take a jQuery object at some point in the future though!

gender404: ah, then `el` makes sense, but I don't think there's a difference between a jQuery Object and a DOM object so I would suggest changing the documentation above to remove jQuery and say DOM object.

jesconstantine: oh good call.  I'll get on that.

jesconstantine: Updated in https://github.com/massgov/mayflower/pull/580/commits/883c0de3ffbe098414f62bf9151345b5fdc94769

jesconstantine:
Got 👍  from @legostud via slack, merging.

PR #567, opened 2017-08-24
DP-3222/3914 Heading level adjustment for accessibility

ygannett:
Heading levels are adjusted from page (node) level templates to their descendant templates by passing the variable `level` or specific variables to propagate proper heading levels.  Templates use default values if such variables don't have their values.

As a result, the heading levels in the page semantically streamed lined through the page.

## Note:
To make this change take effect, below Drupal change needs to be implemented together.

### Related PR for Drupal change:
[DP-3222/3914 Heading level adjustment for accessibility #1157](https://github.com/massgov/mass/pull/1157)

## JIRA
- [DP-3914 [a11y] The headings are not semantically correct](https://jira.state.ma.us/browse/DP-3914)
- [DP-3222 A11y Heading Levels used incorrectly](https://jira.state.ma.us/browse/DP-3222)

### Right rail
right-rail

### Stacked layout
stacked-layout

### Advisory
advisory

### Decision
decision

### Event
event

### Executive order
executive-order

### From
form

### Guide
guide

### How to
how-to

### Location detail
location-detail

### Location
location-office

location-park

### News
news

### Organization landing
organization-landing

### Service detail
service-detail

### Service
service

PR #590, opened 2017-09-13
DP-3222 / 3914 Heading level adjustment for accessibility V3

gender404:
Replacement of [DP-3222/3914 Heading level adjustment for accessibility #567](https://github.com/massgov/mayflower/pull/567)
## Description

Heading levels are adjusted from page (node) level templates to their descendant templates by passing the variable `level` or specific variables to propagate proper heading levels.  Templates use default values if such variables don't have their values.

As a result, the heading levels in the page semantically streamed lined through the page.

## Related Issue / Ticket

- [DP-3914 [a11y] The headings are not semantically correct](https://jira.state.ma.us/browse/DP-3914)
- [DP-3222 A11y Heading Levels used incorrectly](https://jira.state.ma.us/browse/DP-3222)

## Additional Notes:
To make this change take effect, below Drupal change needs to be implemented together.
[DP-3222/3914 Heading level adjustment for accessibility #1157](https://github.com/massgov/mass/pull/1157)


iansholtys (changes requested):
There are some merge conflicts which need to be resolved to merge. Otherwise, I compiled locally and tested all of the page-level templates that were modified.

Things generally worked as expected, but I found three instances of heading levels being skipped. Not sure which of these need to be resolved in this PR and which can be moved to follow-up tickets, but here are the issues either way:
* On the Guide page, “This usually takes approximately 4 weeks” has the wrong heading level. Realistically this doesn’t seem like should be a heading.  This is also an issue towards the top of the How To page.
* On the Guide page, the main headlines under "Moving Day" ("Choose a Reputable Mover" and "Observe Parking Restrictions") are the wrong level. In this instance, it's because the JSON specifies components which are specifically level 4, which can be corrected by calling a different component. Generally speaking, though, a possible solution to this would be to use a level-flexible heading component and merge a level variable into the data passed to the component.
* On the homepage, "Popular Searches" is an h3 when it should be an h2. It looks like the search-banner.twig and helpful-links.twig files haven't been adjusted to be level-flexible.

gender404:
@iansholtys - thanks for the review.  I made my updates based on the issues reported within the two Tickets associated with this PR so I missed the homepage, which doesn't explain the wait time, but like you said it didn't look like a heading so I missed it completely.  

My main goal for this ticket was to provide a `level` on the headings so they could be adjusted as needed and to updated patterns with nested heading to set their child `levels` relative to the parents value.

I've made updates to resolve the issue you found.
1. I replaced the `h5` with a `span` tag instead
2. This is content within a RTE so I changed it to use an `h3`.  We don't really have much control over what the content editors will put in here, but training could help.
3. The problem here was that the Link List (helpful links) component used had it's compHeading set to level: 3 instead of leaving it as the default `h2`

Let me know if you find anything else or if I should be reviewing any other pages.

iansholtys (approved):

iansholtys:
@legostud, as noted on DP-3222, I've tested and the fixes work as expected. To be thorough I checked all pages in the Pages drop down and only saw issues with the "First time?" and "Social" headings at the top of the G2G Service Unemployment Benefits page and Service page. Those aren't in the original scope of the ticket, so this might be ready to go. Either way, I've added more details to DP-3222, and overall this PR seems like a solid improvement to me.

PR #591, opened 2017-09-14
DP-4717 -- Use sentence case for header tags pattern

nstriedinger:
## Description
Remove the css rule for the @molecules/header-tags pattern which currently forces this title case: text-transform: capitalize;

## Related Issue / Ticket

- https://jira.state.ma.us/browse/DP-4517
- [Github issue]()

## Steps to Test

1. Read ticket.
2. Verify that CSS rule has been removed by visiting a page with the relationship indicator.

## Screenshots

## Additional Notes:

Anything else to add?

* 

#### @TODO

*

### DIFF
--- a/styleguide/source/assets/scss/02-molecules/_header-tags.scss
+++ b/styleguide/source/assets/scss/02-molecules/_header-tags.scss
@@ -1,7 +1,6 @@
 .ma__header-tags {
   font-size: 16px;
   letter-spacing: .1em;
-  text-transform: uppercase;
 
   &__terms {
     display: inline;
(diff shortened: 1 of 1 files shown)


gender404 (approved):

styleguide/source/assets/scss/02-molecules/_header-tags.scss
```diff
@@ -1,7 +1,6 @@
 .ma__header-tags {
   font-size: 16px;
   letter-spacing: .1em;
-  text-transform: uppercase;
```
jesconstantine: @nstriedinger does this mean that "more about" is no longer uppercase?  Because I don't think that's the desired affect of this change; we only want the tag terms to have their title case removed.

jesconstantine (changes requested):
@nstriedinger thanks for jumping in there - I think we only need one of your two changes.  Let me know what you think.  Thanks!

jesconstantine (approved):

PR #594, opened 2017-09-15
DP-4534 - Add MF handling for many relationship indicators

nstriedinger:
## Description
Add More/Less dynamic button to "Related To:" items.

## Related Issue / Ticket

- https://jira.state.ma.us/browse/DP-4534
- [Github issue]()

## Steps to Test

1. Ready Ticket
2. Open .json file and add plenty of long "taxonomyTerms" items
3. Visit page and test More/Less functionality.

## Screenshots
Designs are in ticket.

## Additional Notes:

Anything else to add?

* 

#### @TODO

*


styleguide/source/_patterns/02-molecules/header-tags.twig
```diff
@@ -2,9 +2,12 @@
   {% if headerTags.label %}
     <span>{{ headerTags.label }}</span>
   {% endif %}
-  <div class="ma__header-tags__terms">
+  <div id="terms" class="ma__header-tags__terms">
```
jesconstantine: The `id` should be passed in as a variable in case it's used more than once

styleguide/source/_patterns/02-molecules/header-tags.twig
```diff
@@ -2,9 +2,12 @@
...
     {% for item in headerTags.taxonomyTerms %}
       <a href="{{ item.href }}">{{ item.text }}</a>
     {% endfor %}
+    <button type="button" class="js-accordion-link" aria-expanded="false" aria-controls="terms">
```
jesconstantine: The `js` class should be unique to this component.  I'd suggest something like `.js-header-tag-button`

styleguide/source/_patterns/02-molecules/header-tags.twig
```diff
@@ -2,9 +2,12 @@
...
+      <span>more</span><span>less</span>
```
jesconstantine: The button text should be configurable, let's pass them both in as a variable

styleguide/source/assets/js/index.js
```diff
@@ -18,6 +18,7 @@ import pagination        from "./modules/pagination.js";
 import pikaday          from "./modules/pikaday.js";
 import responsiveVideo  from "./modules/responsiveVideo.js";
 import resultsHeading  from "./modules/resultsHeading.js";
+import relationshipIndicators  from "./modules/relationshipIndicators.js";
```
jesconstantine: Could we alphabetize and line these up for readability?

styleguide/source/assets/js/modules/relationshipIndicators.js
```diff
@@ -0,0 +1,44 @@
+export default function (window,document,$,undefined) {
+
```
jesconstantine: Our js module files should use `jQuery.each()` to ensure the code will work with multiple instances of a component and also to ensure that the code only runs when the pattern is present.

styleguide/source/_patterns/02-molecules/header-tags.twig
```diff
@@ -2,9 +2,12 @@
   {% if headerTags.label %}
     <span>{{ headerTags.label }}</span>
   {% endif %}
-  <div class="ma__header-tags__terms">
+  <div id="terms" class="ma__header-tags__terms">
     {% for item in headerTags.taxonomyTerms %}
       <a href="{{ item.href }}">{{ item.text }}</a>
```
jesconstantine: Since you reference this element in your js, we need to use a js class (vs the css class), I'd suggest something like: `.js-header-tag-link`

styleguide/source/assets/js/modules/relationshipIndicators.js
```diff
@@ -0,0 +1,44 @@
...
+  var showHideButton = $('.ma__header-tags__terms .js-accordion-link');
+  var relatedItems = $('.ma__header-tags__terms a');
+  var moreThanThreeRelatedItems = relatedItems.length > 2 ? true : false;
```
jesconstantine: 1. We should use `const` to define variables that don't change (i.e. selectors)
1. `showHideButton` should reference the new js class from prior comment: `.js-header-tags-button`
    1. Since it references a jQuery object, it should also be `$showHideButton`
1. `relatedItems` should reference a js class (vs a css class), I'd suggest something like: `.js-header-tags-link`
    1. Since it references a jQuery object, it should also be `$relatedItems`
    1. Since the only `relatedItems` that we need to worry about are the 3rd+, what if your selector targeted `$('.js-header-tags-link:nth-child(n+3)')`?
1. For `moreThanThreeRelatedItems`:
    1. Including a number in the variable name could be brittle
    1. Since the variable name includes `Three` should you also be testing for `> 3`?
    1. This variable is actually not needed if your $relatedItems variable targets only the items that we ever reveal/hide and if you structure your logic below like:

```
if ($relatedItems.length) {
  showHideButton.hide();
  return; // none of the logic to show / implement the button is necessary
}

// Functionality from your 2nd if statement here
// Init just show 3.
$relatedItems.show()
...

```

styleguide/source/assets/js/modules/relationshipIndicators.js
```diff
@@ -0,0 +1,44 @@
...
+    showHideButton.hide();
+  }
+
+  // If we have related items, inspect them.
+  if (moreThanThreeRelatedItems) {
+    var showThreeRelatedItems = function() {
+      for (var i = 3; i < relatedItems.length; i++) {
+        $(relatedItems[i]).hide();
+      }
+    };
```
jesconstantine: Since your `$relatedItems` selector includes only the items that we want to show/hide, we no longer need this function.

styleguide/source/assets/js/modules/relationshipIndicators.js
```diff
@@ -0,0 +1,44 @@
...
+
+    var showAllRelatedItems = function() {
+      for (var i = 0; i < relatedItems.length; i++) {
+        $(relatedItems[i]).show();
+      }
+    };
```
jesconstantine: Since your `$relatedItems` selector includes only the items that we want to show/hide, we no longer need this function.

styleguide/source/assets/js/modules/relationshipIndicators.js
```diff
@@ -0,0 +1,44 @@
...
+
+    // Init just show 3.
+    showThreeRelatedItems();
```
jesconstantine: This can be `$relatedItems.show()`

styleguide/source/assets/js/modules/relationshipIndicators.js
```diff
@@ -0,0 +1,44 @@
...
+
+    showHideButton.on('click', function(e) {
```
jesconstantine: Please double check to make sure this works for keyboard interaction

styleguide/source/assets/js/modules/relationshipIndicators.js
```diff
@@ -0,0 +1,44 @@
...
+      "use strict";
```
jesconstantine: I think this should be moved to the top of the file (after the `export default` line)

styleguide/source/assets/js/modules/relationshipIndicators.js
```diff
@@ -0,0 +1,44 @@
...
+      e.preventDefault();
+      showHideButton.toggleClass('is-open');
+
+      if (showHideButton.hasClass('is-open')) {
+        showAllRelatedItems();
```
jesconstantine: This can be `$relatedItems.show()`

styleguide/source/assets/js/modules/relationshipIndicators.js
```diff
@@ -0,0 +1,44 @@
...
+        showHideButton.attr('aria-expanded', 'true');
+      }
+      else {
+        showThreeRelatedItems();
```
jesconstantine: This can be `$relatedItems.hide()`

styleguide/source/assets/scss/02-molecules/_header-tags.scss
```diff
@@ -2,6 +2,15 @@
   font-size: 16px;
   letter-spacing: .1em;
   text-transform: uppercase;
+  max-width: 820px;
```
jesconstantine: What's the reasoning for constricting width? Also, if we were going to constrict the width, we should use the [`$l-max-content-width variable`](https://github.com/massgov/mayflower/blob/dev/styleguide/source/assets/scss/00-base/_variables.scss#L12)

styleguide/source/assets/scss/02-molecules/_header-tags.scss
```diff
@@ -27,6 +36,58 @@
...
+    }
+
+    > button {
+      @include ma-button-reset;
+      @include ma-link-underline;
+      @include ma-chevron;
+      padding-left: 10px;
+      font-size: 19px;
+      font-weight: 500;
+      color: #14558f;
```
jesconstantine: Rather than hard code a hex RGB color, this value should be a variable.  See [`$c-theme-primary`](https://github.com/massgov/mayflower/blob/dev/styleguide/source/assets/scss/06-theme/00-base/_colors.scss#L18)

Also any css rules that touch how a pattern will look (i.e. colors, font-size) should be defined in the scss theme directory: https://github.com/massgov/mayflower/tree/dev/styleguide/source/assets/scss/06-theme

styleguide/source/assets/scss/06-theme/03-organisms/_page-header.scss
```diff
@@ -1,7 +1,9 @@
 .ma__page-header {
 
   &__tags {
-
+    @media ($bp-medium-min) {
+      border-color: $c-bd-divider;
```
jesconstantine: Nice!  Thanks for using the variable here!

jesconstantine (changes requested):
@nstriedinger Thanks for your hard work here!  I think there are some areas that we can iterate on.  Let me know if you have any questions.

styleguide/source/assets/js/modules/relationshipIndicators.js
nstriedinger: This is great!

styleguide/source/assets/scss/02-molecules/_header-tags.scss
nstriedinger: The design implied this.

gender404 (approved):

styleguide/source/_patterns/02-molecules/header-tags.twig
```diff
@@ -2,9 +2,12 @@
   {% if headerTags.label %}
     <span>{{ headerTags.label }}</span>
   {% endif %}
-  <div class="ma__header-tags__terms">
+  <div id="{{ headerTags.id }}" class="ma__header-tags__terms js-header-tag-link">
     {% for item in headerTags.taxonomyTerms %}
       <a href="{{ item.href }}">{{ item.text }}</a>
     {% endfor %}
+    <button type="button" class="js-header-tag-button" aria-expanded="false" aria-controls="terms">
+      <span>{{ headerTags.moreLabel }}</span><span>{{ headerTags.lessLabel }}</span>
```
jesconstantine: @gleroux02 what do you think about having these be configurable with reasonable defaults? I'm thinking something like this at the top of the file:

```
{% set moreLabel = headerTags.moreLabel ?: "show more" %}
```

styleguide/source/assets/scss/02-molecules/_header-tags.scss
```diff
@@ -27,6 +36,53 @@
...
+      }
+    }
+
+    > button {
+      @include ma-button-reset;
+      @include ma-link-underline;
+      @include ma-chevron;
+      padding-left: 10px;
+      font-size: 19px;
+      font-weight: 500;
```
jesconstantine: I think `font-weight` needs to go in a `scss/06-theme/02-molecules/_header-tags.scss` file.  

Everything else seems to be in the right place though :)

jesconstantine (commented):
@gleroux02 nice work!  Just one small tiny scss change and a suggestion for button text defaults -- let me know what you think!

styleguide/source/assets/js/modules/relationshipIndicators.js
```diff
@@ -0,0 +1,33 @@
+export default function (window,document,$,undefined) {
+  "use strict";
+  $('.js-header-tag-link').each(function() {
+
+    let $showHideButton = $('.js-header-tag-link .js-header-tag-button'),
+        $relatedItems = $('.js-header-tag-link a:nth-child(n+4)');
+
+      $showHideButton.hide();
```
gender404: This should be done via CSS to avoid a blink cause by JS hiding and showing the button on page load.

styleguide/source/assets/js/index.js
```diff
@@ -18,6 +18,7 @@ import mobileNav                  from "./modules/mobileNav.js";
 import orgSelector                from "./modules/orgSelector.js";
 import pagination                 from "./modules/pagination.js";
 import pikaday                    from "./modules/pikaday.js";
+import relationshipIndicators  from "./modules/relationshipIndicators.js";
```
gender404: you line up the "from" column

styleguide/source/assets/scss/02-molecules/_header-tags.scss
```diff
@@ -27,6 +36,66 @@
...
+      &:after {
+        opacity: .5;
+        margin-left: 0;
+        border-width: 3px;
+        height: 8px;
+        width: 8px;
+        transform: translateY(-45%) rotate(45deg);
+      }
+
+      &.is-open:after {
```
jesconstantine: @gleroux02 I think this selector needs to be updated (just like you did for the spans) so that the caret rotates

styleguide/source/assets/scss/03-organisms/_page-header.scss
```diff
@@ -13,19 +13,28 @@ $page-header-widget-width: 350px;
...
+    margin-bottom: 20px;
 
     .ma__header-tags,
     .ma__social-links {
       border-bottom-width: 1px;
       border-bottom-style: solid;
       min-height: 48px;
       position: relative;
-        top: 1px;
+      top: 1px;
```
jesconstantine: This indentation is intentional to show a dependency between `top` and `position` (a nice little trick from @legostud)

jesconstantine (commented):
@gleroux02 I think once we get the caret rotating this will be good to go!  See my comments below. Thanks!!

jesconstantine (approved):
Woo-hoo!!! Thanks @nstriedinger and @gleroux02 :)

styleguide/source/_patterns/02-molecules/header-tags.twig
```diff
@@ -1,10 +1,16 @@
...
   {% if headerTags.label %}
     <span>{{ headerTags.label }}</span>
   {% endif %}
-  <div class="ma__header-tags__terms">
+  <div id="{{ headerTags.id }}" class="ma__header-tags__terms js-header-tag-link">
     {% for item in headerTags.taxonomyTerms %}
       <a href="{{ item.href }}">{{ item.text }}</a>
     {% endfor %}
+    <button type="button" class="js-header-tag-button" aria-expanded="false" aria-controls="{{ headerTags.id }}">
+      <span>{{ headerTags.moreLabel }}</span><span>{{ headerTags.lessLabel }}</span>
```
powpow12, line 13: this will circumvent line 1 & 2.  @jesconstantine will patch.

styleguide/source/_patterns/02-molecules/header-tags.json
```diff
@@ -1,5 +1,8 @@
 {
   "headerTags": {
+    "id": "terms",
```
powpow12, line 3: this will be done through Js as we do not want to add a new mandatory property without providing some override. @jesconstantine  will provide the JS fix

PR #596, opened 2017-09-15
DP-4313 - Remove logo from service page

nstriedinger:
## Description
Remove optional logo from Service page.

## Related Issue / Ticket

- https://jira.state.ma.us/browse/DP-4313
- [Github issue]()

## Steps to Test

1. Read Ticket
2. View http://localhost:3000/?p=pages-service and verify the removal of logo.

## Screenshots

## Additional Notes:

Anything else to add?

* 

#### @TODO

*

### DIFF
--- a/styleguide/source/_patterns/04-templates/01-content-types/services.json
+++ b/styleguide/source/_patterns/04-templates/01-content-types/services.json
@@ -86,10 +86,10 @@
 
   "introSidebar": {
     "logo": {
-      "alt": "alt text", 
-      "src": "../../assets/images/placeholder/230x130.png",
-      "height": "130",
-      "width": "230"
+      "alt": "",
+      "src": "",
+      "height": "",
+      "width": ""
     },
 
     "social": {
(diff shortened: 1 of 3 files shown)


gender404 (approved):

jesconstantine (changes requested):
@nstriedinger if we're removing the logo from this content type in such a way that it's no longer possible to include it in the future (i.e. removing it from the markup), then we also need to update the corresponding documentation `.md` file and the and the data `.json` file so that future developers understand the scope of the change.

If we want to make the change in a way that would allow us to add the logo back in the future (perhaps not a bad idea - thoughts?) then we should only remove the data sent in the .json file and leave the pattern markup `.twig` and documentation `.md` as is.

nstriedinger:
@jesconstantine It's up to you. What would you like me to do?

jesconstantine:
@nstriedinger let's remove the json data but leave the markup twig and docs md file as is; just make sure the markup only renders the image if the data is sent, and make sure the variable is optional in the md file.  Sound good?

jesconstantine (approved):
Thanks for your updates here @nstriedinger .

PR #599, opened 2017-09-18
DP-4572 - Re-work Dateline

nstriedinger:
## Description
I reworked the Dateline for the Drupal side. FYI -- I kept the `<p>` tag for current PL structure, but we will not be using this pattern on the Drupal end. (On the Drupal end, we're separating everything out and adding the Dateline, and reassembling programmatically.) 

## Related Issue / Ticket

- https://jira.state.ma.us/browse/DP-4572
- [Github issue]()

## Steps to Test

1. Read Ticket
2. go to *?p=pages-press-release*
3. confirm span and bolded dateline.

## Screenshots
screen shot 2017-09-18 at 11 29 29 am

## Additional Notes:

Anything else to add?

* 

#### @TODO

*

### DIFF
--- a/styleguide/source/assets/scss/04-templates/_press.scss
+++ b/styleguide/source/assets/scss/04-templates/_press.scss
@@ -11,17 +11,6 @@
     }
   }
 
-  &__location-label {
-    float: left;
-    line-height: 1em;
-    margin-bottom: 0;
-    padding-top: .25em;
-  }
-
-  &__location-dateline {
-    text-transform: uppercase;
-  }
-
   &__sidebar {
     .ma__contact-list {
 
--- a/styleguide/source/assets/scss/04-templates/_press.scss
+++ b/styleguide/source/assets/scss/04-templates/_press.scss
@@ -20,7 +20,6 @@
 
   &__location-dateline {
     text-transform: uppercase;
-    font-weight: bold;
   }
 
   &__sidebar {
--- a/styleguide/source/assets/scss/04-templates/_press.scss
+++ b/styleguide/source/assets/scss/04-templates/_press.scss
@@ -16,7 +16,11 @@
     line-height: 1em;
     margin-bottom: 0;
     padding-top: .25em;
+  }
+
+  &__location-dateline {
     text-transform: uppercase;
+    font-weight: bold;
   }
 
   &__sidebar {
--- a/styleguide/source/assets/scss/06-theme/03-organisms/_rich-text.scss
+++ b/styleguide/source/assets/scss/06-theme/03-organisms/_rich-text.scss
@@ -2,7 +2,6 @@
 
   &__flame {
     font-weight: bold;
-    text-transform: uppercase;
   }
   
   &__footnote {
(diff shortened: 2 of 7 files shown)


styleguide/source/assets/scss/04-templates/_press.scss
```diff
@@ -16,7 +16,11 @@
     line-height: 1em;
     margin-bottom: 0;
     padding-top: .25em;
+  }
+
+  &__location-dateline {
     text-transform: uppercase;
+    font-weight: bold;
```
gender404: font-weight should be part of a theme CSS file not part of the base CSS file (ie: /06-theme/04-templates/_press.scss file.

gender404 (changes requested):
What was the requirement for adding the extra class to make this bold instead of using the `b` tag?

Personally, I prefer the Drupal solution over what I created.  My solution is a bit brittle so I'd be okay with deleting the `ma__press__location-label` Markup and CSS in favor of just adding the content within the first paragraph of the RTE block.  Having it as part of the RTE block was my original solution, because it was simple and elegant.

styleguide/source/assets/scss/04-templates/_press.scss
nstriedinger: Did I add this correctly?

gender404: yes that looks good

jesconstantine:
@legostud I didn't fully understand your original change request - was it just to move the style rule into the theme scss or was there something else that you thought needed to be addressed here?

nstriedinger:
@legostud -- @gleroux02 suggested I use a `<span>`.

gender404:
This PR doesn't appear to solve any issues so I'm confused as to what's the goal with making this change?  

The `&__label` was already uppercase and the text was already bold.  The only difference I see here is that the &mdash; is no longer uppercase, which doesn't matter.  This PR seems to be just adding bloat for no reason.

gender404:
I feel like we're missing something.  Maybe there is actually more work that needs to be done in Mayflower to match the approach that Mass Gov is using.

gender404:
After talking with Jes, I better understand the desired outcome here.  I would suggest removing all the code for the `location-label` and just add the content `<span class="ma__rich-text__flame">Boston</span> &mdash; ` to the first paragraph of the Rich Text component in Mayflower.  In the _rich-text.scss files we can add the styling needed to make the `&__flame` selector uppercase and bold similar to what you've already created.

What we're doing is adding a 'flame' option to the Rich Text pattern in Mayflower that you can leverage for your current solution.

styleguide/source/assets/scss/06-theme/03-organisms/_rich-text.scss
```diff
@@ -1,4 +1,9 @@
 .ma__rich-text {
+
+  &__flame {
+    font-weight: bold;
```
gender404, line 4: This should be in the 'theme' file for rich-text.  Otherwise this looks really good.

gender404 (changes requested):

styleguide/source/assets/scss/06-theme/03-organisms/_rich-text.scss
gender404: the font-weight: bold that is

gender404 (approved):

PR #604, opened 2017-09-27
DP-3763 Map zoom needs default zoom fix

robyngreen:
## Description
The map zoom on maps with single pins was applying an almost max-level zoom and ignoring any setZoom config overrides. This is a default gmap behavior when a single pin is used and fitBounds is called. See: https://stackoverflow.com/questions/2437683/google-maps-api-v3-can-i-setzoom-after-fitbounds

The fix is to only call fitBounds when more than a single pin is used in `addMarkersToMap`, defaulting to the map config zoom. 

## Steps to Test
1. Visit any page that utilizes a map component and utilizes a single pin. 
2. Observe the default zoom is almost max and cannot be adjusted prior to this PR, but respects the googleMap.js zoom zoom config with it. 

Location / Map components.

#### Today I learned...
https://stackoverflow.com/questions/2437683/google-maps-api-v3-can-i-setzoom-after-fitbounds

This does not have an "elegant" solution to a known problem with gmap and calling fitBounds on single marker maps. All acceptable solutions include either listening for zoom changes or only calling fitBounds conditionally. Ideally the gmap library should probably be patched to either not default zoom on single markers in fitBounds, or allow an optional parameter to specify the zoom level.


jesconstantine (approved):
Thanks for identifying and working around this bug, @robyngreen !  Since this has already passed external review I'm going to squash and merge!

PR #617, opened 2017-10-11
DP-6111: Add initial work for no event state.

gleroux02:
## Description
Adds link and empty text when there are no upcoming events, but past events are available.

## Related Issue / Ticket

- https://jira.state.ma.us/browse/DP-6111

## Steps to Test
1. Go to http://localhost:3000/?p=pages-organization there should be no upcoming events anymore, but the text for "no upcoming events scheduled" should display as well as a link to past events.
2. Go to http://localhost:3000/?p=pages-location-general-content there should be no upcoming events anymore, but the text for "no upcoming events scheduled" should display as well as a link to past events.
3. The organism has been built to show event lists with no upcoming events in sidebar: http://localhost:3000/patterns/03-organisms-by-author-event-listing-past-as-sidebar/03-organisms-by-author-event-listing-past-as-sidebar.html
4. The organism has been built to show event lists with no upcoming events in grid format: http://localhost:3000/patterns/03-organisms-by-author-event-listing-past-as-grid/03-organisms-by-author-event-listing-past-as-grid.html

## Screenshots

## Additional Notes:

Anything else to add?

* 

#### @TODO

*

### DIFF
--- a/styleguide/source/_patterns/03-organisms/by-author/event-listing.twig
+++ b/styleguide/source/_patterns/03-organisms/by-author/event-listing.twig
@@ -13,15 +13,26 @@
       {% include "@atoms/04-headings/sidebar-heading.twig" %}
       {% set teaserHeading = (sidebarHeading.level ? : teaserHeading) + 1 %}
     {% endif %}
-    <ul class="ma__event-listing__items js-event-listing-items">
-      {% for eventTeaser in eventListing.events %}
-        {% set eventTeaser = eventTeaser|merge({"level":teaserHeading}) %}
-        <li class="ma__event-listing__item js-event-listing-item">
-          {% include "@molecules/event-teaser.twig" %}
-        </li>
-      {% endfor %}
-    </ul>
-    {% if eventListing.more %}
+    {% if eventListing.events %}
+      <ul class="ma__event-listing__items js-event-listing-items">
+        {% for eventTeaser in eventListing.events %}
+          {% set eventTeaser = eventTeaser|merge({"level":teaserHeading}) %}
+          <li class="ma__event-listing__item js-event-listing-item">
+            {% include "@molecules/event-teaser.twig" %}
+          </li>
+        {% endfor %}
+      </ul>
+    {% else %}
+      <div class="ma__event-listing__empty">
+        {{ eventListing.emptyText }}
+      </div>
+    {% endif %}
+    {% if eventListing.pastMore and eventListing.pastEvents == true %}
+        <div class="ma__event-listing__past">
+          {% set link = eventListing.pastMore %}
+          {% include "@atoms/11-text/link.twig" %}
+        </div>
+    {% elseif eventListing.more and eventListing.events %}
       <div class="ma__event-listing__more">
         {% set link = eventListing.more %}
         {% include "@atoms/11-text/link.twig" %}
--- a/styleguide/source/_patterns/03-organisms/by-author/event-listing.twig
+++ b/styleguide/source/_patterns/03-organisms/by-author/event-listing.twig
@@ -22,17 +22,19 @@
           </li>
         {% endfor %}
       </ul>
-    {% else %}
+    {% endif %}
+    {% if eventListing.emptyText %}
       <div class="ma__event-listing__empty">
         {{ eventListing.emptyText }}
       </div>
     {% endif %}
-    {% if eventListing.pastMore and eventListing.pastEvents == true %}
-        <div class="ma__event-listing__past">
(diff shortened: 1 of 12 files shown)


styleguide/source/_patterns/03-organisms/by-author/event-listing.twig
```diff
@@ -13,15 +13,26 @@
...
+        {% for eventTeaser in eventListing.events %}
+          {% set eventTeaser = eventTeaser|merge({"level":teaserHeading}) %}
+          <li class="ma__event-listing__item js-event-listing-item">
+            {% include "@molecules/event-teaser.twig" %}
+          </li>
+        {% endfor %}
+      </ul>
+    {% else %}
+      <div class="ma__event-listing__empty">
+        {{ eventListing.emptyText }}
```
jesconstantine: @gleroux02 What do you think about setting a local variable here to pull from `eventListing.emptyText` if populated, and set a reasonable default (i.e. "No upcoming events scheduled") when not?

So this line would become:
```
{{ emptyText }}
```

And at the top of the template we'd have something like:
```
{% set emptyText = eventListing.emptyText ?: "No upcoming events scheduled" %}
```

styleguide/source/_patterns/03-organisms/by-author/event-listing.twig
```diff
@@ -13,15 +13,26 @@
...
+      </div>
+    {% endif %}
+    {% if eventListing.pastMore and eventListing.pastEvents == true %}
```
jesconstantine: I'm curious about checking for both the more link and the presence of the data here in the template.  Part of me feels like that logic decision should be made by the implementation.  (i.e. only send the value for the link if there are past events) . What do you think?

styleguide/source/_patterns/03-organisms/by-author/event-listing.twig
```diff
@@ -13,15 +13,26 @@
...
+        <div class="ma__event-listing__past">
+          {% set link = eventListing.pastMore %}
+          {% include "@atoms/11-text/link.twig" %}
+        </div>
+    {% elseif eventListing.more and eventListing.events %}
```
jesconstantine: Same question as above - part of me thinks maybe checking for the more link is enough and the implementer needs logic to only populate that value when there are events (and if so, how many, etc.).  What do you think?

styleguide/source/assets/scss/03-organisms/_event-listing.scss
```diff
@@ -15,6 +15,15 @@ $event-listing-bp-show-grid: $bp-medium-min;
     @include ma-component-spacing;
   }
 
+  &__empty {
+    font-size: 1.25rem;
```
jesconstantine: We can wait for the design external review, but I think we're going to want this to be `1.5rem` or maybe even `1.65rem` -- cc: @cmajel

jesconstantine (commented):
Thanks @gleroux02 this is off to a great start!!  Thanks for adding the variant/docs!!  I've got a couple questions below.  

Also - can you update the actual organism (http://localhost:3000/?p=organisms-event-listing) docs and data to include the `pastMore`  variable so we can see what it looks like when there are upcoming events, an upcoming more link, and a past more link?  Thanks!

styleguide/source/_patterns/03-organisms/by-author/event-listing.twig
gleroux02: @jesconstantine By setting a default we are then forcing this text to always display, is that what we want? Or should it only display if there are no upcoming events and there are past events.

gleroux02:
@jesconstantine what should this look like if both more links are there? The styling is a bit off currently :)
image

jesconstantine:
@gleroux02 that's a great question - I'll let @cmajel know that we need a design for that.

styleguide/source/_patterns/03-organisms/by-author/event-listing.twig
gleroux02: I don't think I phrased this right. What I should be asking is, should we always be displaying this empty text? Even if there are no past events?

jesconstantine: ah - i gotcha.  yeah no default text makes sense.  and per our slack convo - we'll wrap each piece of content: events, "no events", past link, upcoming link with a logic gate checking for it's existence.  and we'll let the implementer decide the business logic.

gleroux02:
@jesconstantine you can see the logic / documentation updates and font size increase.

jesconstantine (approved):

PR #631, opened 2017-10-16
toggle js-location-listing class at layout breakpoint

clairesunstudio:
## Description
Disable the scroll up to map on mobile on location listing- this feels very jarring, and also does not add any additional info.
## Related Issue / Ticket

- [JIRA issue: DP-5525](https://jira.state.ma.us/browse/DP-5525)

## Steps to Test

- Before: [Mayflower location listing](http://mayflower.digital.mass.gov/?p=pages-location-listing)
- After: [UAT location listing](https://clairesunstudio.github.io/mayflower/?p=pages-location-listing)
1. resize window to 910px breakpoint
2. for window size > 910px, not change
3. for window size < 911px, click/hover on any location listing, check the following:
 - [ ] disable auto scroll to the map
 - [ ] disable active state
 - [ ] disable map update

## Screenshots
location-listing

#### Help needed:
when resize down to window size < 911px, map doesn't recenter.

* N/A

#### @TODO

* N/A


hackajesse (commented):
Passes external review, but needs a changelog @clairesunstudio. Moving to WIP so you can do that. Once you do, move directly to Internal Review for @isaacchansky.

isaacchansky:
Sorry for the super delayed review on this.  I think a simpler way to do this might be to just bind the click/mouseover/etc events once, and check the screen-width upon interaction to either do nothing or continue as normal. Rough code would look like:

```
$el.on('click', function(e) {

  if($(window).width()+15 > 910) {
    // do stuff
  }

});
```

This would avoid the binding/unbinding issues, and simply bind once.

isaacchansky (changes requested):
See previous comment, happy to chat about this on Monday!

PR #634, opened 2017-10-18
Release 5.8.2

isaacchansky:
### Changed
- DP-5858 - Better print styles for accordion components and hiding floating buttons
- DP-5518 - Fix layout bug on feedback form
- DP-5515 - Keep Mayflower content up to date with practical usage in Mass.gov

### Removed
- DP-5636 - On mobile, remove dropshadow between rows on stacked row section.
- DP-5914 - Remove extra padding on callout links

### DIFF
--- a/styleguide/source/assets/js/templates/googleMapInfo.html
+++ b/styleguide/source/assets/js/templates/googleMapInfo.html
@@ -17,7 +17,7 @@
   {{/if}}
   {{#if email}}
     <div class="ma__info-window__email">
-      <span class="ma__info-window__label">Online:</span>
+      <span class="ma__info-window__label">Email:</span>
       <a class="ma__info-window__email-link" href="mailto:{{email}}">{{email}}</a>
     </div>
   {{/if}}
--- a/styleguide/source/assets/js/templates/googleMapInfo.html
+++ b/styleguide/source/assets/js/templates/googleMapInfo.html
@@ -17,7 +17,7 @@
   {{/if}}
   {{#if email}}
     <div class="ma__info-window__email">
-      <span class="ma__info-window__label">Email:</span>
+      <span class="ma__info-window__label">Online:</span>
       <a class="ma__info-window__email-link" href="mailto:{{email}}">{{email}}</a>
     </div>
   {{/if}}
--- a/styleguide/source/assets/scss/02-molecules/_section-links.scss
+++ b/styleguide/source/assets/scss/02-molecules/_section-links.scss
@@ -92,9 +92,12 @@
 
   &__title {
     @include ma-h3;
-    margin-bottom: 15px;
+    margin-bottom: .5em;
+    @media ($bp-small-min){
+      margin-bottom: .5em;
+    }
     align-self: stretch;
-
+    
     a {
       border: none;
     }
@@ -166,6 +169,7 @@
     @include ma-reset-list;
   }
 
+
   // skip the first item
   &__item + &__item {
     margin-top: 1em;
@@ -173,13 +177,14 @@
 
   // Make child content callout links full width
   &__item > .ma__callout-link {
-    @include ma-container();
     display: block;
     width: 100%;
   }
 
   &__item > .ma__decorative-link {
+    &__item > .ma__decorative-link {
     font-size: 1.625rem;
     line-height: 1.3;
+  }
   }
 }
--- a/styleguide/source/assets/scss/02-molecules/_section-links.scss
+++ b/styleguide/source/assets/scss/02-molecules/_section-links.scss
@@ -94,6 +94,7 @@
     @include ma-h3;
     margin-bottom: .5em;
     @media ($bp-small-min){
+      // override ma-h3
       margin-bottom: .5em;
     }
     align-self: stretch;
--- a/styleguide/source/assets/scss/02-molecules/_section-links.scss
+++ b/styleguide/source/assets/scss/02-molecules/_section-links.scss
@@ -97,7 +97,7 @@
       margin-bottom: .5em;
     }
     align-self: stretch;
-    
+
     a {
       border: none;
     }
@@ -182,9 +182,7 @@
   }
 
   &__item > .ma__decorative-link {
-    &__item > .ma__decorative-link {
     font-size: 1.625rem;
     line-height: 1.3;
-  }
   }
 }
--- a/styleguide/source/assets/scss/03-organisms/_stacked-row-section.scss
+++ b/styleguide/source/assets/scss/03-organisms/_stacked-row-section.scss
@@ -1,11 +1,12 @@
 .ma__stacked-row-section {
 
   @media ($bp-large-max) {
-    box-shadow: 0 0.5rem 0.5rem -0.25rem rgba(1, 1, 1, 0.25);
+    box-shadow: none;
     padding-bottom: 45px;
 
-    &:last-child {
-      box-shadow: none;
+    & + & {
+      border-top-width: 1px;
+      border-top-style: solid;
     }
   }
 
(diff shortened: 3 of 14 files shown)


styleguide/source/assets/js/templates/googleMapInfo.html
```diff
@@ -17,7 +17,7 @@ <h3 class="ma__info-window__name">{{{name}}}</h3>
   {{/if}}
   {{#if email}}
     <div class="ma__info-window__email">
-      <span class="ma__info-window__label">Email:</span>
+      <span class="ma__info-window__label">Online:</span>
       <a class="ma__info-window__email-link" href="mailto:{{email}}">{{email}}</a>
```
jesconstantine: Seems like we need some better logic here.  I see the fix to make the label more general, but we hard code `mailto` in the link because we are clearly expecting an email address.  I'm guessing that means any non-email data sent here are actually broken.  Would it make more sense to have a dedicated link for email and a dedicated link for web/online?

@isaacchansky What do you think is the scope of fixes that is appropriate for a release branch?

styleguide/source/assets/scss/02-molecules/_section-links.scss
```diff
@@ -1,185 +1,190 @@
...
+    width: 100%;
+  }
+
+  &__item > .ma__decorative-link {
+    &__item > .ma__decorative-link {
+    font-size: 1.625rem;
+    line-height: 1.3;
+  }
+  }
+}
```
jesconstantine: It's hard to tell what actually changed here, @isaacchansky are you confident that whatever has changed is patch level?  Also - I think we need an .editorconfig for this project, what do you think?

jesconstantine (commented):
@isaacchansky I think this is probably okay - but see my two comments and let me know what you think.

Also, I noticed some js stuff was broken from your github pages (I noticed because there were no map markers appearing on maps for me), but when I deployed the release to mine things seem fine.  So I wonder if there is still a kink to work out of in the deploy script.

styleguide/source/assets/scss/02-molecules/_section-links.scss
```diff
@@ -1,185 +1,190 @@
...
+  }
+
+  // Make child content callout links full width
+  &__item > .ma__callout-link {
+    display: block;
+    width: 100%;
+  }
+
+  &__item > .ma__decorative-link {
+    &__item > .ma__decorative-link {
```
gender404: This is broken.  Maybe a bad merge happened.

~~~
&__item > .ma__decorative-link {
 +    &__item > .ma__decorative-link {
~~~

translates to `.ma__section-links__item > .ma__decorative-link .ma__section-links__item > .ma__decorative-link__item > .ma__decorative-link`

gender404: if you fix the line ending problem (switch to unix or windows line endings) then you will be able to see what was actually changed.

styleguide/source/assets/js/templates/googleMapInfo.html
isaacchansky: @jesconstantine good call, I'll hit up Rob and see if there's a good decision to be made here quickly, or we can maybe revert/note a fix for later if it seems innocuous for now?

styleguide/source/assets/scss/02-molecules/_section-links.scss
isaacchansky: mhm! Added, let me know what you think, I just set the minimal options for line endings for now.

styleguide/source/assets/js/templates/googleMapInfo.html
isaacchansky: Ok I think drupal actually handles this all itself, so this is just a change to keep consistent with that? either way, let's talk more about it, I've reverted the merge for this change
.

styleguide/source/assets/scss/02-molecules/_section-links.scss
```diff
@@ -92,7 +92,10 @@
 
   &__title {
     @include ma-h3;
-    margin-bottom: 15px;
+    margin-bottom: .5em;
+    @media ($bp-small-min){
+      margin-bottom: .5em;
+    }
```
gender404: I feel I mentioned this before.  why are we overriding the base style with the base style?  Is this an override of the ma-h3 mixin?  if so a comment would be helpful.

styleguide/source/assets/scss/03-organisms/_stacked-row-section.scss
```diff
@@ -1,11 +1,12 @@
 .ma__stacked-row-section {
 
   @media ($bp-large-max) {
-    box-shadow: 0 0.5rem 0.5rem -0.25rem rgba(1, 1, 1, 0.25);
+    box-shadow: none;
```
gender404: are we overriding something?  Not sure why we need to have `box-shadow: none` if we're never using a box shadow.

styleguide/source/assets/scss/04-templates/_stacked-row.scss
```diff
@@ -4,11 +4,12 @@
   &__section {
 
     @media ($bp-large-max) {
-      box-shadow: 0 0.5rem 0.5rem -0.25rem rgba(1, 1, 1, 0.25);
+      box-shadow: none;
```
gender404: same as above - why `none` verses deleting the property

styleguide/source/assets/scss/06-theme/03-organisms/_stacked-row-section.scss
```diff
@@ -1,6 +1,6 @@
 .ma__stacked-row-section {
 
-  & ~ & &__container:before {
+  &, & ~ & &__container:before {
```
gender404: ?? Each Stacked Row Section gets a border-color and then each container after the first gets a border color as well.  This probably isn't causing anything to break, because it's dependent on the base SCSS setting a border width and style, but it seems very odd and could cause some future issues.

~~~
.ma__stacked-row-section,
.ma__stacked-row-section ~ .ma__stacked-row-section .ma__stacked-row-section__container:before {
~~~

styleguide/source/assets/js/templates/googleMapInfo.html
jesconstantine: Thanks for reverting - Drupal only populates variables `{{ }}` so we should definitely revisit this.

styleguide/source/assets/scss/02-molecules/_section-links.scss
isaacchansky: yeah, after looking into it, that's definitely an override. just commented it.

gender404:
@isaacchansky - be careful with doing reverts.  I'm not sure what will happen once a reverted change is merged back into dev.  Will it revert the change there as well?

isaacchansky:
@legostud yeah the revert will apply to dev when it's merged, so all PRs I'm reverting will have to get re-rolled and adjusted

gender404:
@isaacchansky - Could you delete this release branch and start over with some cherry picks to avoid the reverts?

isaacchansky:
@legostud sure thing

styleguide/source/assets/scss/03-organisms/_stacked-row-section.scss
clairesunstudio: @legostud, you are right. have looked into your comments and these are not overriding things so removed in a [pr 635](https://github.com/massgov/mayflower/pull/635) thanks for flagging these!

styleguide/source/assets/scss/06-theme/03-organisms/_stacked-row-section.scss
clairesunstudio: hey @legostud, so I found out that this is needed because the rule is the border of`.ma__stacked-row-section`
screen shot 2017-10-19 at 4 48 51 pm

see [PR #637](https://github.com/massgov/mayflower/pull/637)

PR #636, opened 2017-10-19
Dp 6106 new topic banner mobile

clairesunstudio:
## Description

Design fixes:
Translucent green fade banner becomes full width, but color, fade, and transparency remain the same as they were before, and the image stays in the background.
Icon remains on desktop view, but is removed for mobile.
Title is vertically aligned to center, and remains aligned left in the same position relative to left margin as it was before.
Decrease banner height to 330px to accommodate up to 5 text lines.
See more in Jira ticket DP-6106

## Related Issue / Ticket

- [JIRA issue](https://jira.state.ma.us/browse/DP-6106)
- [Github issue]()

## Steps to Test

- **Before:** [mayflower topic page](http://mayflower.digital.mass.gov/?p=pages-topic)
- **After:** [UAT topic page](https://clairesunstudio.github.io/mayflower/?p=pages-topic)
- [x] fix the gap on the right when window width < 480px 
- [x] vertical center align banner content < 480px 
- [x] remove banner icon at breakpoint ~780px 

## Screenshots
topic-banner

upper limit current in the CMS:
screen shot 2017-10-19 at 2 05 36 pm

## Additional Notes:

Anything else to add?

* 

#### @TODO

*

### DIFF
--- a/styleguide/source/assets/scss/03-organisms/_page-banner.scss
+++ b/styleguide/source/assets/scss/03-organisms/_page-banner.scss
@@ -1,9 +1,10 @@
 $page-banner-icon-width: $column + $gutter;
+$page-banner-height: 330px;
 
 .ma__page-banner {
   background-size: cover;
   background-position: center center;
-  height: 400px;
+  height: $page-banner-height;
   margin-bottom: 20px;
   overflow: hidden;
   position: relative;
@@ -133,10 +134,10 @@ $page-banner-icon-width: $column + $gutter;
   }
 
   &--columns &__content {
-    height: 400px;
+    height: $page-banner-height;
     max-width: 640px;
     position: relative;
-      top: 200px;
+      top: calc(#{$page-banner-height}/2);
 
     @media ($bp-medium-min) {
       height: 100%;
@@ -312,7 +313,7 @@ $page-banner-icon-width: $column + $gutter;
   &--columns &__image {
     background-position: 50% 50%;
     background-size: cover;
-    height: 400px;
+    height: $page-banner-height;
     position: absolute;
       right: 0;
       top: 0;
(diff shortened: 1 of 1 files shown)


styleguide/source/assets/scss/03-organisms/_page-banner.scss
```diff
@@ -133,10 +134,12 @@ $page-banner-icon-width: $column + $gutter;
...
 
   &--columns &__content {
-    height: 400px;
+    height: $page-banner-height;
     max-width: 640px;
     position: relative;
-      top: 200px;
+      top: calc(#{$page-banner-height}/2);
+      left: 20px;
+      margin-left: -20px;
```
gender404: wouldn't this be `top: 50%` ?  Add `transform: translateY(-50%);`  to make it truly centered.

Also why not `left: 0` ?

clairesunstudio: hey @legostud, setting` top:50%` with `translate` didn't work. so I had to explicitly set the top this way. setting `left` is to fix the gap on the right. please let me know if you see a better way to fix this. thanks!
screen shot 2017-10-19 at 2 27 21 pm

gender404: I would think just adding `width:100%;` would get rid of the gap on the right.

clairesunstudio: let me try that!

clairesunstudio: so width:100% didn't work because the parent container had a padding. fixed it by removing the padding from container instead

hackajesse (approved):
Passes external review. Still issues with topic page banners, but that's for another ticket. Moving to Internal Review and assigning to @isaacchansky .

isaacchansky (approved):

PR #641, opened 2017-11-07
bump up page banner height on larger screens

isaacchansky:
## Description
Increase height of service page banner on desktop to account for longer page titles.

## Related Issue / Ticket

- [DP-5388](https://jira.state.ma.us/browse/DP-5388)

## Steps to Test

1. View service page: `http://localhost:3000/?p=pages-service` 
2. Height of banner on desktop should be 400px instead of 300px. 

## Screenshots
screen shot 2017-11-07 at 12 59 01 pm

## Additional Notes:

This seems like a band-aid solution to some degree, since longer content will always overflow, and a flexible height banner with consistent padding _may_ make more sense? But, I'm not sure we want or pragmatically care about a flexible height banner given the actual nature of the current content. 

* Page Banner on desktop

#### @TODO

*

### DIFF
--- a/styleguide/source/assets/scss/03-organisms/_page-banner.scss
+++ b/styleguide/source/assets/scss/03-organisms/_page-banner.scss
@@ -22,7 +22,7 @@ $page-banner-icon-width: $column + $gutter;
 
   &--overlay {
     @media ($bp-large-min) {
-      height: 300px;
+      height: 400px;
     }
   }
 


hackajesse (approved):
Passes external review. Good suggestion.
@cmajel does Isaac's suggestion merit a ticket to re-evaluate the front end implementation?

iansholtys (approved):
Works as expected. As a side-note I do agree with Isaac; I don't see a compelling reason to use a fixed height. A minimum height could make sense from a stylistic perspective but given the variation in the length of permissible headlines, I'm not sure trying to design a one-size-fits-all makes sense here.

hackajesse:
Design team will take a look a this and other similar banner height issues in next/ other cycles. We'll merge this in as a sufficient fix for the moment.

PR #646, opened 2017-11-21
DP-6916 Prevent hyphen in "learn more about this organization" link in contact group

iansholtys:
## Description
The contact group's link hyphenation rule was removed to prevent auto-hyphenation. As no other hyphenation rule applies to the link by default, simple removal of the rule was adequate to fix this behavior.

## Related Issue / Ticket
https://jira.state.ma.us/browse/DP-6916

## Steps to Test

1. Check out branch locally and compile.
2. Visit http://localhost:3000/?p=pages-howto
3. Use inspector to modify the web link under the Contact sidebar block. It should be changed to "Learn more about this great organization".
4. Confirm that the link does not break on the word "organization".

## Screenshots
hyphens

## Additional Notes:

N/A

* A subset of the Contact Group molecules where link text is present and long enough to wrap onto multiple lines

#### @TODO
N/A

#### Today I learned...
N/A

### DIFF
--- a/styleguide/source/assets/scss/02-molecules/_contact-group.scss
+++ b/styleguide/source/assets/scss/02-molecules/_contact-group.scss
@@ -63,7 +63,6 @@
     display: inline-block;
 
     a {
-      hyphens: auto;
       overflow-wrap: break-word;
       -ms-word-break: break-all;
       word-break: break-all;


hackajesse (approved):
@iansholtys nice work. Passes ER and moves to IR w/ @isaacchansky - 
Question, though - when I changed the copy using the Inspector, and then went from desktop to mobile, the copy reverted back to the original. I changed it on mobile to something and widened into desktop view and the copy changed back to my original change from the first time I did desktop. Any reason for this? Looks like it was two distinct sets of copy almost...

isaacchansky:
@hackajesse You're correct in thinking that there are two pieces of markup that swap. The reason for that is so on mobile contact info get hoisted to the top of the page, rather than bumped to the bottom. This CSS fix applies to both though, so we're all good 👍

PR #651, opened 2017-11-30
DP-5817 rules of court

jaymejohnson:
This PR is for work on the Rules of Court content type and related elements. It should be considered a WIP PR until updated to say otherwise. 

The definition of done can be found here: https://jira.state.ma.us/browse/DP-5817

## Steps to Test

QA steps, when they exist will be written in the original ticket, here: https://jira.state.ma.us/browse/DP-5817


styleguide/source/_patterns/04-templates/01-content-types/court-rules.md
```diff
@@ -0,0 +1,69 @@
...
+* Jump Links
+* Rich Text
+* Form Downloads
+* Footnote list
+* Header Tags
+* Press Listing
+* Event Listing
+* This template contains [Twig Blocks](https://twig.symfony.com/doc/2.x/tags/extends.html) that can be used to populated the Page Content, Right Rail, or Post Content sections with patterns found in Mayflower
+
+### JavaScript Used
```
jesconstantine: @jaymejohnson - We should mention the javascript for adding heading anchor tags and also heading indentation here

styleguide/source/_patterns/04-templates/01-content-types/court-rules.twig
```diff
@@ -0,0 +1,90 @@
...
+        {% include "@organisms/by-author/contact-list.twig" %}
+      {% endif %}
+    </div>
+  </div>
+  <div class="ma__court-rules__content">
+    {% if jumpLinks %}
+      {% include "@organisms/by-template/jump-links.twig" %}
+    {% endif %}
+    <div class="main-content main-content--two">
+      <div class="page-content js-insert-heading-anchors js-outline-indent">
```
jesconstantine: I believe we were talking about scoping these 2 js classes at the `@organisms/by-author/rich-text` level as opposed to hard coding them here in the template.  Maybe it could look like 2 new variants of the rich text organism which each key off of a new boolean property that we'll be sure to set for the court rules content type and page patterns.

That way the rich text js module can also be cleaner and not have to crawl up the DOM for a parent class.

styleguide/source/_patterns/05-pages/court-rules~with-indentation.json
```diff
@@ -0,0 +1,742 @@
+{
+  "pageHeader": {
+    "category": "Rule of Court Category",
+    "subCategory": {
+      "compHeading": {
+        "title": "with Indentation",
```
jesconstantine: What do you think about using this data structure for the court-rules.json page data instead of adding this page variant / version?

styleguide/source/assets/js/helpers/headerAnchorLinks.js
```diff
@@ -0,0 +1,29 @@
+
+export default function (window,document,$,undefined) {
+
+  $('.js-insert-heading-anchors').each(function(){
```
jesconstantine: If we scope this class to the rich text organism, should this JS live there as well?  Do you think there's value in having this class applied to other patterns?  I can't think of many that would need it because most will have an id... thoughts?

styleguide/source/assets/js/modules/richText.js
```diff
@@ -4,5 +4,15 @@ export default function (window,document,$,undefined) {
     let $el = $(this);
 
     $el.find('table').wrap( "<div class='ma__rich-text__table-wrapper'></div>" );
+
+    // Allow heading and contents to be indented.
+    if ($el.parents('.js-outline-indent').length) {
```
jesconstantine: If we make this class scoped to rich text, then we can just check for this class on the element itself which is cleaner.

jesconstantine (commented):
@jaymejohnson - off to a great start here!  I think my comments here address some of your questions from slack.

I also think we an remove any work re: accordion-content and collapsible-header -- that was just POC code that I was playing around with, nothing that should be deployed.  I also see some work related to other page types here - curious what that's about?

Let me know if you have any questions.  Thanks!

jaymejohnson:
> I also think we an remove any work re: accordion-content and collapsible-header -- that was just POC code that I was playing around with, nothing that should be deployed.

Oops. I thought that those were new patterns at first, so they slipped past me. (Which was easy to do since I went about incorporating your POC work in totally the wrong way. I'd have done it differently had I known)

>  I also see some work related to other page types here - curious what that's about?

I have no idea! Hahahaha. It came in with merges. I thought if I pulled down the upstream and merged it into my feature branch this other stuff would go away, but it didn't...

jaymejohnson:
@jesconstantine I am making a To Do list here, the goal of which is having this done and ready to go. Please feel free to add anything you think I am missing. I am sure I am missing things, there are lots of moving parts here. One thing in particular I am concerned about is forgetting to update MD files.

- [x] Get rid of the extra stuff concerning other content types that got pulled into here.

- [x] Remove any POC code, e.g. accordion-content and collapsible-header.

- [x] Remove Indentation sample page variant, and just use default page template to show indentation.

- [x] Mention the javascript for adding heading anchor tags and also heading indentation in styleguide/source/_patterns/04-templates/01-content-types/court-rules.md

- [x] Scope `js-insert-heading-anchors js-outline-indent` classes to the `@organisms/by-author/rich-text` level

- [x] `styleguide/source/assets/js/modules/richText.js` lines 7-9: Once the ` js-outline-indent` class is scoped to the rich-text organism and is no longer in any content type templates, we can just check for this class on the element itself instead of looking for it's parents as it is currently set up.

styleguide/source/_patterns/05-pages/court-rules~with-indentation.json
jaymejohnson: I thought it felt redundant to have both pages anyway. I'm adding this to my checklist below.

styleguide/source/assets/js/helpers/headerAnchorLinks.js
jaymejohnson: If we could know for certain that we would never be able to use it anywhere else, I'd say scope it to the rich-text organism. But for now, the idea of having it at least available to other patterns if we would need it seems better than having a future developer have to reverse engineer what we did and pull it out again anyway. That's my thinking on it right now, anyway.

jaymejohnson:
This is ready for review, @jesconstantine. :)

styleguide/source/_patterns/03-organisms/by-author/rich-text.twig
```diff
@@ -1,4 +1,6 @@
-<section class="ma__rich-text js-ma-rich-text">
+{% set headerIndent =  richText.headerIndent ? 'js-insert-heading-anchors js-outline-indent' : '' %}
```
jesconstantine: @jaymejohnson are you having the `richText.headerIndent` property add both a `js-outline-indent` and the `js-insert-heading anchors` classes intentionally here?  I think we'll need a separate property and separate local variable for the anchor tags functionality.

Also - please update the documentation `.md` file associated with this pattern with the new variable(s) and their usage guidelines.

styleguide/source/_patterns/03-organisms/by-template/page-header.twig
```diff
@@ -27,6 +27,12 @@
         {{ pageHeader.category }}
       </div>
     {% endif %}
+
+    {% if pageHeader.subCategory %}
+      {% set compHeading = pageHeader.subCategory.compHeading %}
```
jesconstantine: @jaymejohnson this seems like an instance where we will always know what level of heading we want to use - what do you think?  could we set a default value here in the twig?

styleguide/source/_patterns/03-organisms/by-template/page-header.md
```diff
@@ -23,6 +23,8 @@ This pattern is adds an `<h1>` page title and other helpful bit of content that
 pageHeader: {
   category: 
     type: string / optional,
+    subcategory:
```
jesconstantine: Please bump this indentation out to be a sibling to `category` and please also update the capitalization to `subCategory`

styleguide/source/_patterns/03-organisms/by-template/page-header.md
```diff
@@ -23,6 +23,8 @@ This pattern is adds an `<h1>` page title and other helpful bit of content that
...
+    type: string / optional,
```
jesconstantine: I think this type should be `Comp Heading` instead of `string` right?

styleguide/source/_patterns/03-organisms/by-template/page-header~for-court-rules.json
```diff
@@ -0,0 +1,93 @@
+{
+  "pageHeader": {
+    "category": "Rule of Court Category",
+    "subCategory": {
+      "compHeading": {
```
jesconstantine: I'm not sure we need to retain the `compHeading` property - could we namespace it with `subCategory` like you have here but just have the properties of a `compHeading` nested directly under?  That seems to be the pattern that we follow - so long as we document this in the `.md` file and we update the variable assignment accordingly.

styleguide/source/_patterns/03-organisms/by-template/page-header~for-court-rules.json
```diff
@@ -0,0 +1,93 @@
...
+                  "text": "Amended July 17, 2007 effective October 1, 2007"
+                },{
+                  "text": "Amended June 2, 2003 effective October 18, 2003"
+                }]
+              }
+            ]
+          }
+        }
+      },{
+        "path": "@atoms/11-text/paragraph.twig",
```
jesconstantine: I think the pattern that [we've established for similar instances of this description text](https://github.com/massgov/mayflower/blob/dev/styleguide/source/_patterns/04-templates/01-content-types/policy-advisory.json#L38) is to have this paragraph text contained in a rich text organism, so the data structure would be: `pageHeader.optionalContents[].path` set to the rich text twig and `pageHeader.optionalContents[].data` set to the `richText` data structure which would contain this paragraph `path` and `data` here.

styleguide/source/_patterns/03-organisms/by-template/page-header~for-court-rules.md
```diff
@@ -0,0 +1,8 @@
+### Description
+This is a variant of the [Page Header](./?p=organisms-page-header) pattern showing an example of how it is being used on a Court Rules page.
+
+### How to generate
+* Set the `publishState` variable to null (if it's not already)
+* Replace the `category` text with the text you want
+* Set subCategory to null if no subCategory is desired, otherwise set the compHeading title.
```
jesconstantine: Thanks for adding these docs!  I think we want more than just the title populated, please update this documentation line to something like: "Set the `subCategory` variable to null (if no sub category is needed) or populate it with the variables for a `compHeading`"

styleguide/source/_patterns/04-templates/01-content-types/court-rules.json
```diff
@@ -0,0 +1,476 @@
...
+                  "text": "Amended July 17, 2007 effective October 1, 2007"
+                },{
+                  "text": "Amended June 2, 2003 effective October 18, 2003"
+                }]
+              }
+            ]
+          }
+        }
+      },{
+        "path": "@atoms/11-text/paragraph.twig",
```
jesconstantine: Just noting another occurrence of this data structure where we want to contain this paragraph in a rich text organism.

styleguide/source/_patterns/04-templates/01-content-types/court-rules.md
```diff
@@ -0,0 +1,71 @@
...
+* Contact List (js/modules/accordions.js)
+* Footnote List (js/modules/footnote.js)
+* Jump Links (js/modules/scrollAnchors.js)
+* Anchor tags (js/helpers/headerAnchorLinks.js)
```
jesconstantine: Can we add a little more context here by saying something like: "Anchor tags for headings within rich text"

styleguide/source/_patterns/04-templates/01-content-types/court-rules.twig
```diff
@@ -0,0 +1,90 @@
...
+      {% include "@organisms/by-template/jump-links.twig" %}
+    {% endif %}
+    <div class="main-content main-content--two">
+      <div class="page-content">
+        {% for richText in mainContent.contents %}
+          {% include "@organisms/by-author/rich-text.twig" %}
+        {% endfor %}
+
+        {% block pageContent %}
+          {#{% include "@base/placeholder.twig" with {'placeholder':{'text':'Page Content'}} %}#}
```
jesconstantine: Let's leave this placeholder uncommented so implementers know there's a block in which they can add other patterns.

styleguide/source/_patterns/05-pages/rules-of-court.twig
```diff
@@ -0,0 +1,11 @@
+{% extends '@templates/01-content-types/court-rules.twig' %}
+
+{% block header %}
+  {% include "@organisms/by-template/header.twig" %}
+{% endblock %}
+
```
jesconstantine, line 6: Please add an empty block for page content here: `{% block pageContent %}{% endblock %}`

(See https://github.com/massgov/mayflower/blob/dev/styleguide/source/_patterns/05-pages/policy-advisory-directive.twig#L7 for an example)

styleguide/source/assets/js/helpers/headerAnchorLinks.js
jesconstantine: I'm not sure we'd need this for headings outside of rich text (because most heading patterns have `id` variables.  So I think we should move this js into rich text js - what do you think?

I also think we need to add a separate local variable `headingAnchors` to the rich text twig template which is empty by default but can be populated with the appropriate js class when a config boolean value on rich text data structures is set to true: `richText.headingAnchors`

What do you think?

styleguide/source/assets/js/helpers/headerAnchorLinks.js
```diff
@@ -0,0 +1,29 @@
...
+      // For H3+
+      if ($(heading).prop("tagName") !== 'H2') {
+        // Create an id based on the heading text.
+        let id = slugify($(heading).text());
+        // Insert an anchor tag before the heading.
+        $(heading).before("<a name='"+id+"'></a>");
+      }
+    });
+
+    function slugify(text)
```
jesconstantine: Please move this function to it's own helper file

styleguide/source/assets/js/modules/richText.js
```diff
@@ -4,5 +4,15 @@ export default function (window,document,$,undefined) {
     let $el = $(this);
 
     $el.find('table').wrap( "<div class='ma__rich-text__table-wrapper'></div>" );
+
+    // Allow heading and contents to be indented.
```
jesconstantine: Please add a little more context to this comment which mentions the relationship between this js and the scss

styleguide/source/assets/scss/03-organisms/_rich-text.scss
```diff
@@ -171,10 +171,50 @@ $rich-text-spacing-mobile: 1.5rem;
...
   & > :last-child {
     margin-bottom: 0 !important;
   }
-}
+
+  // Indent courts types rich text content as outline
+  .ma__court-rules__content & {
+
+    &__indent[data-ma-heading-parent='H2'] {
+      margin-left: 20px;
```
jesconstantine: Please set a local sass variable here set to the value of `$gutter` (see: https://github.com/massgov/mayflower/blob/dev/styleguide/source/assets/scss/00-base/_variables.scss#L3) and use that instead of hard coding pixels in this line and in the following lines.

jesconstantine (changes requested):
This is great work @jaymejohnson !!  I wrote several small comments that I think we can breeze through together and then this will be good to go!  Let me know how I can help.  Thanks :)

styleguide/source/assets/js/helpers/headerAnchorLinks.js
jaymejohnson: I agree. This has been changed and pushed.

styleguide/source/_patterns/03-organisms/by-author/rich-text.twig
jaymejohnson: You are right, they are definitely better separated. I guess I was thinking that there probably wouldn't be a need for the one without the other. At any rate, this has been fixed and pushed. An update to the `rich-text.md` file has also been pushed up.

styleguide/source/_patterns/03-organisms/by-template/page-header.md
jaymejohnson: Fixed.

styleguide/source/_patterns/03-organisms/by-template/page-header.md
jaymejohnson: Fixed.

styleguide/source/_patterns/03-organisms/by-template/page-header~for-court-rules.json
jaymejohnson: Fixed.

styleguide/source/_patterns/03-organisms/by-template/page-header.twig
jaymejohnson: @jesconstantine 
I changed it to this:
```{% if pageHeader.subCategory %}
      {% set compHeading = pageHeader.subCategory %}
      {% include "@atoms/04-headings/comp-heading.twig" %}
    {% endif %}
```

Is that what you had in mind? And then I changed the data structure for
`page-header~for-court-rules.json` as follows:

```
 "subCategory": {
        "title": "Rule of Court Optional Sub-category",
        "level": 3,
        "color": "",
        "id": "",
        "centered": ""
    },
```

styleguide/source/_patterns/03-organisms/by-template/page-header~for-court-rules.json
jaymejohnson: See above comment.

styleguide/source/_patterns/03-organisms/by-template/page-header~for-court-rules.md
jaymejohnson: Fixed.

styleguide/source/_patterns/04-templates/01-content-types/court-rules.json
jaymejohnson: I think I fixed this when I fixed the others.

styleguide/source/_patterns/04-templates/01-content-types/court-rules.md
jaymejohnson: Fixed. Also, I changed the path to reflect the fact that functionality for this lives in `rich-text.js `

styleguide/source/_patterns/04-templates/01-content-types/court-rules.twig
jaymejohnson: Fixed. I restored all the placeholders in that file while I was at it.

styleguide/source/_patterns/05-pages/rules-of-court.twig
jaymejohnson: Fixed.

styleguide/source/assets/js/helpers/headerAnchorLinks.js
jaymejohnson: Fixed.

styleguide/source/assets/scss/03-organisms/_rich-text.scss
jaymejohnson: Fixed.

styleguide/source/_patterns/03-organisms/by-template/page-header.twig
jesconstantine: For the part about restructuring the data that's exactly what I meant.

For the part about having the `pageHeader` be opinionated with a `subCategory.level` default value, I meant setting a default in twig (which could be overridden in json - by implementers).

I just pushed up a commit with this update: https://github.com/massgov/mayflower/pull/651/commits/d75080c255219d6a33782eb81d3e1f11c0799a92

styleguide/source/assets/scss/03-organisms/_rich-text.scss
jesconstantine: Nice work!  

I just realized we were hard coding a scope (to court rules content type) for the ability to use these js / css hooks for outline indentation.  Fixed in https://github.com/massgov/mayflower/pull/651/commits/6501f09ee7ae700001e7a9cf7db7ce5aced9516e and then refactored to fold into existing heading styles in https://github.com/massgov/mayflower/pull/651/commits/2eefa5f31457a7270f6bd83b197348c46a4be718

styleguide/source/_patterns/02-molecules/listing-table.twig
```diff
@@ -4,7 +4,49 @@
       {% for row in listingTable.rows %}
         <tr>
           <th scope="row">{{ row.label }}</th>
-          <td class="ma__rich-text">{{ row.text|raw }}</td>
+          {# Backwards compatible with 5.8.x where listingTable.rows array item objects expect label and text properties instead of a label and an "items" array. #}
```
jesconstantine: @gleroux02 Just wanted you to see that this new show/hide functionality of the listing table pattern is non-breaking.  We do require a new data structure to use the new functionality, but the old data structure will still work just as it always did.

styleguide/source/_patterns/03-organisms/by-author/rich-text.md
```diff
@@ -26,6 +28,13 @@ This pattern displays the contents of a rich text editor with an optional title
 ### Variables
 ~~~
 richText: {
+
+  headerIndent: {
+    type: boolean / optional
+  },
+  anchorLinks: {
+    type:boolean / optional
+  },
```
jesconstantine: @gleroux02 in order to get the anchor links and outline indentation for the rules of court content type, when you pass the `richText` data structure for the necessary fields, set these new variables to true.

jesconstantine (approved):
@jaymejohnson IMO this is ready to go!

Wondering if you or @gleroux02 (whoever has time!) could just take a look at my 3 commits on top of your branch which I mentioned in our comment dialogue:

RE: letting page header pattern be opinionated about what level that compHeading pattern should be (but also overridable): https://github.com/massgov/mayflower/commit/d75080c255219d6a33782eb81d3e1f11c0799a92

RE: Allowing any page type to implement rich text with outline indentation: https://github.com/massgov/mayflower/commit/6501f09ee7ae700001e7a9cf7db7ce5aced9516e and refactored to fold into existing rich text heading styles (and to resolve a sass compile error) https://github.com/massgov/mayflower/commit/2eefa5f31457a7270f6bd83b197348c46a4be718

Once either of you is happy with those commits, we can squash and merge!!

jaymejohnson:
Looks good. Let's ship it!

PR #654, opened 2017-12-07
Add new gulp-copy dependency to fix npm-install + gulp build

jesconstantine:
## Description

This PR fixes the build issue where gulp-copy is not installed after `npm-install` despite its being required in a new gulp task.

## Related Issue / Ticket

n/a

## Steps to Test

1. Pull this branch down + check it out
1. `cd` into the `styleguide` directory
1. Run `npm install`
1. Run `gulp`
1. Verify that you are able to build Mayflower locally

## Screenshots
Error that was appearing after running `npm install` after  running the command `gulp` would time out after ~19 seconds
error from mayflower

## Additional Notes:

Anything else to add?

* Build system

#### Today I learned...

If your mayflower work includes installing a new node package, you have to run `npm shrinkwrap --dev` to add your new dependency to the `npm-shrinkwrap.json` file and you have to commit that file (so committing the package.json is not enough).  See: https://docs.npmjs.com/cli/shrinkwrap

### DIFF
--- a/styleguide/npm-shrinkwrap.json
+++ b/styleguide/npm-shrinkwrap.json
@@ -3593,6 +3593,12 @@
         }
       }
     },
+    "gulp-copy": {
+      "version": "1.0.1",
+      "from": "gulp-copy@latest",
+      "resolved": "https://registry.npmjs.org/gulp-copy/-/gulp-copy-1.0.1.tgz",
+      "dev": true
+    },
     "gulp-debug": {
       "version": "2.1.2",
       "resolved": "https://registry.npmjs.org/gulp-debug/-/gulp-debug-2.1.2.tgz",


isaacchansky (approved):
👍 thanks for putting this one together!

mrossi113 (approved):
👍 It worked! I was able to run `gulp` after `npm install`

mrossi113:
Screenshot of gulp working!!! @jesconstantine Thank you!

screen shot 2017-12-07 at 4 14 58 pm

gender404:
@jesconstantine - why do you need gulp-copy?  That's kind of what Gulp does by default with src and dest.

jesconstantine:
@legostud I thought I needed `gulp-copy` to get the `prefix` option (https://www.npmjs.com/package/gulp-copy#options).  Is there a way I can accomplish the same thing just through the gulp task config?  Below is the [gulp task](https://github.com/massgov/mayflower/blob/dev/styleguide/tools/gulp/tasks/copyTwig.js) that I set up, for context.

The crux of the issue is that we are serving `.twig` files so that we can make an ajax request to them (for client side twigjs rendering).  Pattern Lab knows that directory as `_patterns` but for whatever reason github pages doesn't like the `_` so we need the served directory to be named `patterns`.  Totally open to other / easier ways to accomplish this same end result.

```
module.exports = function copyTwigTask(config, env){

    // copy files settings
    var twig = {
        src: [
            config.patternLabRoot + "/source/_patterns/**/*.twig"
        ],
        dest: "public/assets/patterns"
    };

    // register the watch
    quench.registerWatcher("copyTwig", twig.src);


    /* copy files */
    gulp.task("copyTwig", function(next) {

        return gulp.src(twig.src, { base: config.root })
            .pipe(quench.drano())
            .pipe(copy(twig.dest,{prefix: 2}))
            .pipe(gulp.dest(config.dest))
            .pipe(debug({title: "copyTwig:"}));
    });
};
```

gender404:
@jesconstantine: something like this:
https://github.com/massgov/mayflower/blob/dev/styleguide/tools/gulp/tasks/svg2twig.js

jesconstantine:
@legostud thanks I'll give that a shot and yank gulp-copy if it works!

PR #667, opened 2018-01-04
DP-7193 wider layout for contact accordions

isaacchansky:
## Description

Contact list accordions' contact groups only stack, so on wider widths this looks funny. This PR adds a flexbox layout to the content within contact-me accordions, which works at any accordion width.  
Notably, this attempts to make a really stable layout which works when the accordion is placed into any width container.

## Related Issue / Ticket

- [JIRA issue](https://jira.state.ma.us/browse/DP-7193)

## Steps to Test

1. View comp: https://projects.invisionapp.com/share/VNEV7ARM5#/screens/268281226
2. Run Mayflower locally 
3. Ensure that things look right at each any viewport width (note: comp only displays desktop)
4. Ensure that this does not mess with other content groups at all viewport widths
    - http://localhost:3000/patterns/05-pages-executive-order/05-pages-executive-order.html 
    - http://localhost:3000/patterns/05-pages-howto/05-pages-howto.html
    - http://localhost:3000/patterns/05-pages-site-alert/05-pages-site-alert.html

## Screenshots
contact-list

* molecule/contact-us
* organism/contact-list

#### @TODO

*

### DIFF
--- a/styleguide/source/assets/scss/02-molecules/_contact-us.scss
+++ b/styleguide/source/assets/scss/02-molecules/_contact-us.scss
@@ -63,6 +63,10 @@
     transform: rotate(135deg);
   }
 
+  &--accordion &__content {
+    padding: 0 20px 30px;
+  }
+
   .js &--accordion &__content {
     display: none;
   }
@@ -75,20 +79,20 @@
     justify-content: flex-start;
     flex-basis: 25%;
 
-    // width overlaps parent's 1px borders
+    // width overlaps parent's 1px borders & negates parent padding
     // This is so we can have divider borders (left border) and since we have a wrapping
     // flexbox grid, the first item on each show needs its left border hidden.
-    width: calc(100% + 2px);
+    width: calc(100% + 42px);
     position: relative;
-      left: -1px;
+      left: -21px;
   }
+
   &__content-wrap .ma__contact-group {
     max-width: 25%;
     min-width: 18rem;
     padding-left: 30px;
     padding-right: 30px;
     margin-top: 20px;
-    margin-bottom: 40px;
     border-left: 1px solid #dcdcdc;
 
     // Reset if we aren't in an accordion
@@ -111,6 +115,11 @@
     }
 
 
+  }
+
+  &__content-wrap .ma__contact-us__more-info {
+    padding-left: 20px;
+    padding-right: 20px;
   }
 
   &__extra {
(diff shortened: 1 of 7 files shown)


tnathanjames (approved):
Looks good to me.

styleguide/source/assets/scss/02-molecules/_contact-us.scss
```diff
@@ -63,18 +63,54 @@
     transform: rotate(135deg);
   }
 
-  &--accordion &__content {
-    padding: 0 20px 30px;
```
iansholtys: Removal of these padding rules is causing issues for the organism in general at mobile resolutions (there's no padding at the bottom) and for implementations which have a more info link at all resolutions. A screenshot of the latter is attached below.



It doesn't appear to be broken on http://www.isaacchansky.com/mayflower/patterns/05-pages-howto/05-pages-howto.html but I'm not convinced this has been properly updated; rules (such as this one) which are clearly removed from the PR are still showing up in the Inspector.

Related to the refactoring of margins and padding, there's substantially more vertical space between Contact List items on the How To page at a bunch of resolutions (one example, when the window is ~1200px wide) and it looks a bit off to me.

iansholtys (changes requested):
Found some margin/padding quirks.

Also, while I couldn't find a contextual place to post the comments, margin/padding is off when Contact Lists are used in sidebars. At desktop resolutions they're fine, but when the sidebar becomes full-width and contact items go into columns, things aren't quite right. There's no space between the individual items, and second+ items have padding on top which causes them to be misaligned.

isaacchansky:
@iansholtys I refactored a bit and think got things working in a better way - If the contact list in the sidebar is still an issue can you provide a link to where you're seeing it? thanks!

iansholtys:
@isaacchansky, your changes resolve a lot of issues I had with margin and padding, with the exception of the contact list when used in the sidebar. You can see it on the Site Alert Page at mobile and desktop resolutions: http://localhost:3000/?p=pages-site-alert

Screenshot attached.

iansholtys:
@isaacchansky, that takes care of the issues at desktop resolution, though when the sidebar goes full-width under the main content and the Contact Groups are side-by-side, the `.ma__contact-group + .ma__contact-group {padding-top: 20px}` is causing them to be misaligned. That said, if there are >2 items and they wrap onto a second line, having space between them is important... the best fix I've been able to come up with is to use padding- or margin-top on each contact group, then using the same amount of negative margin-top on `.ma__contact-us__content-wrap` to offset it. Maybe I'm missing an easier fix, though.

iansholtys (approved):

iansholtys:
All the issues I noticed have been resolved.

PR #677, opened 2018-01-25
DP-7359 Rich text tables rendering differently in Chrome vs IE

ygannett:
## Description

Change the `word-break` value for links in rich text, `.ma__rich-text a`,  to be consistent to all browsers instead of Chrome specific.  The original value `word-break: break-word;` only applies to WebKit/Blink. 

All browsers wrap a word in links from rich text instead of breaking down to characters in Chrome.  The change only applies to WebKit/Blink based browsers. No changes to other browsers since their style is set to what's expected.

## Related Issue / Ticket

- [DP-7359 Rich text tables rendering differently in Chrome vs IE](https://jira.state.ma.us/browse/DP-7359)

## Steps to Test

Note:  You can test in Mayflower by editing one of rich text output in sample pages.  I don't find any link samples in rich text section in Mayflower.  Or, use the tag, `5.10.1-alpha-7359`, to test in Drupal. If you test in Drupal, you can use the exiting page which already has a link in a table cell in rich text field, /service-details/dor-legal-library.

Test items:
- a link in rich text section, `.ma__rich-text a`
- a link in a table cell in rich text section, `.ma__rich-text table th a` and `.ma__rich-text table td a`    

1. Check the style change.  In **Inspect** or **Firebug**, find `word-break: normal;` for the test item, `.ma__rich-text a` and  `.ma__rich-text table th a` or `.ma__rich-text table td a` in any browsers as shown in the screenshot **After** below.

2. Check visual.  To check visual, the link needs to be displayed in the smaller parent container than the longest word in the link. It is easier/quicker to test with the sample page mentioned above in Drupal.  You could also set up your sample to meet the condition.  In Chrome, find the test item is display a whole word without breaking down.
Note: Make sure you test in Chrome, which shows breaking down words in the current production environment as shown in the screenshot **Before** below.  Other browsers in the current production environment show the test item like the screenshot **After**.

## Screenshots
**Before**
dp-7359_before

**After**
dp-7359_after

## Additional Notes:

Anything else to add?

* 

#### @TODO

*

### DIFF
--- a/styleguide/source/assets/scss/03-organisms/_rich-text.scss
+++ b/styleguide/source/assets/scss/03-organisms/_rich-text.scss
@@ -192,7 +192,7 @@ $heading-indent: $gutter;
     overflow-wrap: break-word;
     transition: border .3s;
     word-wrap: break-word;
-    word-break: break-word;
+    word-break: normal;
   }
 
   // end link styles


rbayliss (approved):
Looks good to me!

PR #688, opened 2018-02-05
DP-2405 Change the container for an abbreviation in a page banner title to <abbr> from <span>

ygannett:
## Description

Add semantics to the organization name abbreviation by replacing `<span>` with `<abbr>`.
The screen reader support for `title` attribute with `<abbr>` is not great at this point. Since the abbreviation is presented right after the full name of the organization, it omits `title` attribute with its value of full organization name, and  the supplement info. "also known as" before the tag is added as screen reader only text.  
Screen readers announce the `<h1>` as "_organization full name_ also known as _abbreviation_".

## Related Issue / Ticket

- [DP-2405 [a11y] Not using \<abbr\> for DFG](https://jira.state.ma.us/browse/DP-2405)

## Steps to Test

1. Go to an organization page.
2. Locate the page banner title `h1` in source code.
3. Find the organization name's abbreviation is:
       - wrapped with `<abbr>` instead of `<span>`.
       - no visual change
4. Find `span.visually-hidden` is added with its content "also known as " between the title text and `<abbr>`.
5. Test the area with VoiceOver to see it's read as "_organization full name_ also known as _abbreviation_".

## Screenshots

## Additional Notes:

Anything else to add?

* 

#### @TODO

*

### DIFF
--- a/styleguide/source/_patterns/03-organisms/by-template/page-banner.twig
+++ b/styleguide/source/_patterns/03-organisms/by-template/page-banner.twig
@@ -41,7 +41,7 @@
       <h1 class="ma__page-banner__title">
         {{ pageBanner.title }}
         {% if pageBanner.titleSubText %}
-          <span> {{ pageBanner.titleSubText }}</span>
+          <abbr title="{{ pageBanner.title }}"> {{ pageBanner.titleSubText }}</abbr>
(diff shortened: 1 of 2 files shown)


yriahi (approved):
LGTM

PR #714, opened 2018-02-28
DP-6883 Determine feasibility of embedding SVGs 1x per page

ygannett:
Any PRs being created needs a changelog.txt file before being merged into dev. See: [Change Log Instructions](/docs/change-log-instructions.md)

## Description

Set up a system to pull SVG from multiple places in a page and the site from once source per output.

### Round 1

-  Set up one global source for 
     - the SVG icons used in the global part (header and footer)
     - the social media icons
     - the PDF icon for the test case

- Test for their efficiency.

MF alpha tag: `5.14.0-alpha-68833`
Corresponding Drupal PR: 

#### Mar 2nd Note
Test branch:  DP-6883_svg-optimization-op1 - REMOVED (3/5)

With this branch, set up a generic template, `@atom/05-icons/svg.twig`, for svg icon generation.
The `icon` property is replaced with a icon name instead of a unique icon twig template under `@atom/05-icons`.

```
{% set assetsPath = url.assetsPath %}
{% set iconPath = '/' ~ assetsPath ~ '/images/icon-' ~ icon ~ '.svg#icon-' ~ icon %}
<svg aria-hidden="true">
  <use xlink:href="{{ iconPath }}"></use>
</svg>
```
The template is called in its parent template:
```
{% include "@atoms/05-icons/svg.twig" %}
```

The icon source files are set up under `/assets/images/`.  Their file name starts with `icon-` followed by _icon name_.svg.

The first line, `{% set assetsPath = url.assetsPath %}`, cannot be a part of the template to work in Drupal.  Drupal doesn't define and use `url.assetsPath`.

Next step is set up a function with an argument of _icon_ to generate an icon in the template.
Then, this template is not necessary anymore.  The function could be called in its parent template level.

In the function, `assetPath` is defined for Mayflower use only.
In Drupal, another function with the same function name is set up to do its own part.

#### Mar 5th Note
The test branch has been merged into this branch.
No sprite with this test.  All icon files are set up individually.
@sumeetpareek helped me to find the assets path commonly used between Mayflower and Drupal.
Setting up a simple template to assess the efficiency.

#### To take advantage of external source:
- an icon is used multiple times in the page
- an icon is used multiple times in the site when a user visit multiple pages
- some icons used multiple times in the same patterns can be grouped to one file as sprite to reduce http request

#### To take advantage of inline source:
If an icon is used only in a specific page, it's better to leave it as inline.
Ex) Topic pages, Transition page

### Round 2
Analyze: 
- which icons are better to be external 
- which icons can be grouped together to reduce http requests

MF alpha tag: `5.14.0-alpha-68834`
Corresponding Drupal PR: 

## Related Issue / Ticket

- [DP-6883 Determine feasibility of embedding SVGs 1x per page](https://jira.state.ma.us/browse/DP-6883)

## Steps to Test

1. 

## Screenshots

## Additional Notes:

Anything else to add?

* 

#### @TODO

*


gender404:
What's the goal here?  We originally started this project with all SVG icons as part of an SVG sprite that was generated by Gulp.  

We moved away from that for a few reasons:
1. Some of the icons needed to be uploaded by content authors so we could add those icons to the sprite.
2. The sprite file became very large (might not be an issue here if you're going to have multiple sprite files).
3. The sprite file required a request per page load.  Being a large file that delayed the rendering of the page.  With multiples sprite files the individual requests can be quicker, but you'll end up with more request overhead.

The reason we moved to inline SVG files - converted to Twig with the Gulp task `gulp svg2twig` (which will blow away all this work if it is run again):
1. Allows the SVG code to be part of the html so it can be cached
2. Eliminates the request required to load the SVG

rbayliss:
Hey @legostud,
The goal is to reduce the HTML size of pages that reuse the same SVG many times.  Eg: https://www.mass.gov/lists/all-audit-reports (this one page is 8MB).  These pages are slow to download and do DOM parsing on, and very costly to store in the backend caches.  In these situations, it's preferable to the user to make one external request rather than waiting for a large HTML download to finish, and it's preferable for our backend caches not to store the SVG at all.  It sounds like the implementation isn't quite right yet since it modifies files that are generated automatically, but I think that's ok for now - this is still experimental.

gender404:
@rbayliss - is there a way to compress the html?  I know with CSS files doing gzip on repeating patterns reduces it quite a bit.

gender404:
Wow that page is extreme.  A sprite is probably a good idea there.  You should be able to re-enable the gulp svg-sprite code to build them for you automatically.  I would then suggest building a few sprite files based on common types, like downloads and social media.  If you can make separate folders for each svg-sprite that will make it easier for gulp to build it and will prevent the svg2twig task from overriding them.

The other option is to use an image tag with src equal to the svg.  You would have to put the fill color back on the path of the SVG, but that is probably fine for this type of icon that, I think, is only used here.

rbayliss:
@legostud, that HTML is served as GZIP when downloaded, so for most users, the actual transfer size is nowhere near 8mb.  There's no way to compress it in our backend caches though, unfortunately, and I think the DOM performance concerns are still pretty real regardless of compression (although we haven't figured out how use is gonna affect that). Thanks for the tip on gulp `svg-sprite`.  We'll take a look at that.

PR #779, opened 2018-04-16
DP-8579: update padding for social links

clair0917:
DP-8579 -- Social links on elected org page overlap executive area when the links are long or more than x number

## Description
Adds padding tot he right to keep the social links from overlapping the person card

## Related Issue / Ticket

- [JIRA DP-8579](https://jira.state.ma.us/browse/DP-8597)

## Steps to Test
github pages is still down
test with the organization details / elected official page

## Screenshots
screen shot 2018-04-16 at 12 55 41 pm


clair0917:
Note: no backstop images to include with this PR, because no visual changes included.

isaacchansky:
I'm seeing the padding but it seems to be blowing out the right side instead of forcing the items to wrap. I think we might need a `flex-wrap: wrap`? 

screen shot 2018-04-18 at 1 46 39 pm

styleguide/source/assets/scss/06-theme/03-organisms/_about-section.scss
```diff
@@ -72,6 +72,10 @@
         @media ($bp-small-min) {
           display: flex;
```
isaacchansky, line 73: I think we need a `flex-wrap:wrap` on this for the padding to work

clair0917:
These updates were moved into https://github.com/massgov/mayflower/pull/792

clair0917:
https://github.com/massgov/mayflower/pull/792

PR #780, opened 2018-04-16
DP-8516 - Add flex basis style to CSS that was applied inline.

edwardchan:
Any PRs being created needs a changelog.txt file before being merged into dev. See: [Change Log Instructions](https://github.com/massgov/mayflower/blob/dev/docs/change-log-instructions.md)

## Description
Added the `flex-basis: 100px` style to the CSS for the bio pages that was previously only added as inline styles in the page pattern. This update should have that style applied to the integration piece in DP-8516.

## Related Issue / Ticket
https://jira.state.ma.us/browse/DP-8516

## Steps to Test

[ ] Go to view the bio page - /?p=pages-bio-page.
[ ] Inspect element on the organization image. You should see that the `flex-basis: 100px` style is added and applied to the image:
image
 

#### @TODO

*

### DIFF
--- a/changelogs/DP-8516.txt
+++ b/changelogs/DP-8516.txt
@@ -1,0 +1,4 @@
+___DESCRIPTION___
+Changed
+Minor
+- DP-8516: Moved inline style to SCSS file.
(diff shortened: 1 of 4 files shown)


isaacchansky (approved):
Looks good to me 👍 

Edit: actually, I'm confused why this is necessary - unless we're removing the inline style, won't that still override the CSS?

edwardchan:
Closing this PR since passing in the "width" from the integration piece gets the inline style to show up.

PR #787, opened 2018-04-18
DP-8645  remove nbsp from decorative links

clair0917:
## Description
Removes the non-breaking space from the decorative links and uses css to space correctly 

## Related Issue / Ticket

- [DP-8645](https://jira.state.ma.us/browse/DP-8645)

## Steps to Test

1.  Visit the Organization elected officials page  (/?p=pages-organization-elected-official)
2. Inspect one of the decorated links
3. You should see that the nbsp has been removed from the link

## Screenshots
screen shot 2018-04-18 at 5 18 20 pm

## Additional Notes:
There are a large number of backstop files included, because removing the space and refactoring the css caused some links to break on different words.


isaacchansky (changes requested):
I think the non-breaking-space initially was required to avoid having the arrow icon become an orphan on narrower widths of decorative links. I'll attach a gif to explain what I mean more specifically, I'm not sure there's another way to do this without a non-breaking-space.

isaacchansky:
decorative-link-line-break

PR #792, opened 2018-04-19
DP-8597  social links on elected page overlap update

clair0917:
DP-8579 -- Social links on elected org page overlap executive area when the links are long or more than x number

## Description
Adds padding tot he right to keep the social links from overlapping the person card

## Related Issue / Ticket

- [JIRA DP-8597](https://jira.state.ma.us/browse/DP-8597)

## Steps to Test
github pages is still down
test with the organization details / elected official page

## Screenshots
screen shot 2018-04-16 at 12 55 41 pm


isaacchansky (approved):
👍 Looks great to me!

PR #806, opened 2018-04-27
Sync master dev

mrossi113:
#804


rbayliss:
We're going to merge this one now, as it's just a sync of master back to dev.  If there are any issues with these reference screenshots, they should be resolved in the dev branch before Weds.

rbayliss (approved):

PR #807, opened 2018-04-30
DP-8363 Reduce the front end performance impact of the main logo and state seal

ygannett:
## Description

- Break down the site logo in the header with the state seal image and the text, Mass.gov.
- For the seal, use the same file as one in the footer.
- Resize the footer seal image from 240 x 240 to 200 x 200, which is minimum required size for HiDPi/Retina display.  The file size has been reduced 5kb.

Associated Drupal PR:  [DP-8363 Reduce the front end performance impact of the main logo and state seal #2224](https://github.com/massgov/mass/pull/2224)

## Related Issue / Ticket

- [DP-8363 Reduce the front end performance impact of the main logo and state seal](https://jira.state.ma.us/browse/DP-8363)

## Steps to Test

1. Go to pages with difference display patterns with the site logo (see the screenshots below), and compare any difference between production.

readme2_desktop

homepage_desktop

homepage_mobile

## Screenshots

## Additional Notes:

Anything else to add?

* 

#### @TODO

*


styleguide/source/_patterns/01-atoms/09-media/site-logo.twig
```diff
@@ -1,5 +1,6 @@
 <div class="ma__site-logo">
   <a href="{{ url.domain }}/" title="Mass.gov home page">
-    <img src="{{ url.domain }}/{{ url.assetsPath }}/images/mass-logo.png" alt="Mass.gov" width="164" height="75" />
+    <img src="{{ url.domain }}/{{ url.assetsPath }}/images/stateseal.png" alt="Mass.gov" width="164" height="75" />
```
rbayliss: @ygannett - I think the width and height attributes here are out of date, right?  We should update them to the correct dimensions, as this is the size that's picked up on initial render (before the CSSOM is constructed).  Setting them correctly should prevent jitter as the logo loads in. I think we can safely set it to the desktop size (45/45), as that's the majority of our traffic.

ygannett: Yes, that needs to be updated as well.  Also, the size attributes are causing skewed display in IE, so removed the `width` attribute.  Browsers still able to figure out the display size from the height attribute in case CSS is not rendered.

rbayliss: Ok, we can review the attributes in IR, once design review has passed.

rbayliss (approved):

PR #801, opened 2018-04-25
DP-8712  support 2 executives in about row

clair0917:
## Description
Adds a variation of the about section with 2 cards

## Related Issue / Ticket

- [JIRA issue DP-8712](https://jira.state.ma.us/browse/DP-8712)

## Steps to Test

1. The new pattern variation can be seen at /?p=organisms-about-section-two-officials

## Screenshots
screen shot 2018-04-25 at 12 23 13 pm

## Additional Notes:
No backstop files added because nothing changed

PR #816, opened 2018-05-04
DP-8825 Add expandable teaser listing for org boards

isaacchansky:
## Description

Adds a variation of the teaser listing organism to allow for an accordion layout.

## Related Issue / Ticket

- [JIRA issue](https://jira.state.ma.us/browse/DP-8825)

## Steps to Test

1. Run maflower locally
2. Ensure than the existing listing table organism works and looks as it used to
3. Look at the organization boards page, and see that it matches these invision links:
    - https://invis.io/6WEP0CK7M#/295217761_orgboardclosed
    - https://invis.io/6WEP0CK7M#/295217760_orgboardopen 

## Additional Notes:

Anything else to add?

* Added `@organisms/by-author/teaser-listing~as-expandable

#### @TODO

* Maybe add blocks around each item subsection?


isaacchansky:
@tnathanjames is this an instance where we would want two blocks around each sub-section of `items` in the twig file?

tnathanjames:
It looks like these blocks are the same, it is just the number of items that changes?  If that is the case, it might make sense to provide something like we did in this PR: https://github.com/massgov/mayflower/pull/801/files, printing the block that was previously defined.

clair0917 (approved):
looks good

PR #824, opened 2018-05-09
Release 5.22.0

mrossi113:
### Added
- DP-6715: Allow main content of image-promos to be overwritten.

### Changed
- DP-7761: Take html out of map pin title.
- DP-8363: Replace the site logo image with the seal image + text.
- DP-8877: MF create front end code to support display of short description on org page that is separate from contact row.
- DP-8880: Allow a block to be overridden for the sidebar of the stacked row section pattern.
- DP-8891: re-orders Mayflower elements for the Org page variations.


mrossi113:
Tested all the PRs and passed.

PR #839, opened 2018-05-16
Release 5.24.0

mrossi113:
### Fixed
- DP-8956: Update contact us to provide a way to keep address above other information even if there is only two contact groups.


mrossi113:
Passed the smoke test.

PR #840, opened 2018-05-16
Sync Master back to dev from 5.24.0

mrossi113:
#839


mrossi113:
Syncing master back to dev from release 5.24.0

PR #849, opened 2018-05-23
DP-6907  change default ordered list type styling in rich text fields

clair0917:
## Description
Changes the ol structure to correctly wrap nested ordered lists in list items in Mayflower, to match the correct WYSIWYG layout
Adds a few extra nested lists to the pattern
Adds custom list style type style

## Related Issue / Ticket

- [JIRA issue DP-6907](https://jira.mass.gov/browse/DP-6907)

## Steps to Test
1. Visit /?p=atoms-ordered-list to see the new ordered list style types 

## Screenshots
screen shot 2018-05-23 at 2 09 26 pm

## Notes
The styles are written to accommodate up to 9 nested lists, after which all counters will be digits


isaacchansky (approved):
The nesting is a little gnarly but it's obvious what is going on so I like it 👍

PR #853, opened 2018-05-29
DP-8897  / DP-9077  relationship indicators to expose organizations on public pages  conflict free

clair0917:
## Description
Adds relationship indicators pattern and updates org page with pattern for review

## Related Issue / Ticket

- [DP-8897](https://jira.state.ma.us/browse/DP-8897)
- [DP-9077](https://jira.mass.gov/browse/DP-9077)

## Steps to Test

1. The pattern can be viewed here: /?p=molecules-relationship-indicators
2. The pattern has also been added to the org page here: /?p=pages-organization

## Screenshots
screen shot 2018-05-16 at 10 39 36 am
screen shot 2018-05-29 at 12 19 55 pm


styleguide/source/assets/js/modules/headerTags.js
```diff
@@ -30,6 +30,47 @@ export default function (window,document,$,undefined) {
...
+      let $tagStateText = $tagState.text();
+
+      $tagWrapper.parent().toggleClass('tags-open');
+      $button.toggleClass('is-open');
+      $tagState.text($tagStateText === 'fewer' ? 'more' : 'fewer');
+      $hiddenTag.toggle();
+    });
+
+    $(window).resize(function () {
+      // remove all the screen width specific styles
```
isaacchansky, line 61: Where are widths being set that they need to be removed?

styleguide/source/assets/js/modules/headerTags.js
```diff
@@ -30,6 +30,47 @@ export default function (window,document,$,undefined) {
...
+      $buttonCounter.removeAttr('style');
+      $hiddenTag.removeAttr('style');
+
+      // recount the hidden tags and update the button text
+      setTimeout(function(){
```
isaacchansky, line 66: Why the timeout?

styleguide/source/assets/js/modules/headerTags.js
clair0917: the number of tags being shown is different for mobile and desktop and the counter in the button is based on the hidden tags. For example, on mobile, all the tags for 'offered by' are closed on mobile, but a default of one is shown on desktop. Clicking the show more button toggles the tags open. When you change the screen width, they should revert back to the original styles so the buttons still open correctly.

styleguide/source/assets/js/modules/headerTags.js
clair0917: So that the style attributes would finish updating _before_ the tag counter ran again. I was having trouble getting the tag count to change consistently. Half the time the layout would switch to desktop, the single default tag would show, but the button still said "show 2 more"
Adding a half second pause allowed for the count to update correctly

isaacchansky:
cool, comments make sense

isaacchansky (approved):

PR #866, opened 2018-06-20
Release 5.29.0

mrossi113:
### Added
- DP-9146/DP-8825: Adds teaser listing group organism to apply accordion behavior to multiple teaser listings at once.

### Changed
- DP-9243: MF fix info details to show contact and related in sidebar as well as bottom on desktop.

### Removed 
- DP-1234: Removed alt value from the state seal since the text "Mass.gov" is duplicate to the content in the followed <span>.


mrossi113:
LGTM ER all three ticket!

